PR #1, opened 2021-04-10
Add draft CI workflow

BrunoGrandePhD:
I'm not sure what's the best way of testing these CI workflows other than reviewing them and making sure they behave as expected once merged. 

I've opted for a setup where we have two stack groups: `dev` and `prod`. The `dev` stacks will be deployed to the `workflows-nextflow-dev` account whereas the `prod` stacks will be deployed to the `workflows-nextflow-dev` and `workflows-nextflow-prod` accounts. Another pattern I noticed [here](https://github.com/Sage-Bionetworks/iAtlas-infra/tree/staging/config) is to have `dev`, `prod`, and `common`. 

You can ignore the change affecting the `CODEOWNERS` file. I just removed the extra line at the end of the file to satisfy the pre-commit test.

### DIFF
--- a/.github/workflows/aws-deploy.yaml
+++ b/.github/workflows/aws-deploy.yaml
@@ -16,7 +16,7 @@ jobs:
       - uses: pre-commit/action@v2.0.0
 
   sceptre-deploy:
-    if: ${{ github.event_name == 'push' }}
+    if: github.event_name == 'push'
     runs-on: ubuntu-latest
     needs: pre-commit
     steps:
@@ -43,26 +43,34 @@ jobs:
           WORKON_HOME: ~/.local/share/virtualenvs
           PIPENV_CACHE_DIR: ~/.local/share/pipcache
         run: pipenv install --dev
-      - name: Login as AWS Service User (Under Dev Account)
+      - name: Assume AWS Role (Dev Account)
         uses: aws-actions/configure-aws-credentials@v1
         with:
           aws-access-key-id: ${{ secrets.CI_SERVICE_USER_ACCESS_KEY }}
           aws-secret-access-key: ${{ secrets.CI_SERVICE_USER_SECRET_ACCESS_KEY }}
           aws-region: us-east-1
-      - name: Deploy both 'dev' and 'prod' stack groups to 'dev' account using sceptre
+          role-to-assume: arn:aws:iam::035458030717:role/SceptreBuildAccessRole
+          role-duration-seconds: 1200
+      - name: Deploy 'dev' and 'prod' Stack Groups to Dev Account
         run: |
+          pipenv shell
           sceptre launch dev --yes
           sceptre launch prod --yes
+
+      # Commented out until the AWS account becomes available. Track this PR:
+      # https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/135
+
       # - name: Assume AWS Role (Prod Account)
       #   uses: aws-actions/configure-aws-credentials@v1
-      #   if: ${{ github.ref == 'refs/heads/main' }}
+      #   if: github.ref == 'refs/heads/main'
       #   with:
       #     aws-access-key-id: ${{ secrets.CI_SERVICE_USER_ACCESS_KEY }}
       #     aws-secret-access-key: ${{ secrets.CI_SERVICE_USER_SECRET_ACCESS_KEY }}
       #     aws-region: us-east-1
-      #     role-to-assume: arn:aws:iam::0987654321:role/NextflowWorkflows-CIServiceRole
+      #     role-to-assume: arn:aws:iam::0123456789:role/SceptreBuildAccessRole
       #     role-duration-seconds: 1200
-      # - name: Deploy both 'prod' stack groups to 'prod' account using sceptre
-      #   if: ${{ github.ref == 'refs/heads/main' }}
+      # - name: Deploy 'prod' Stack Group to Prod Account
+      #   if: github.ref == 'refs/heads/main'
       #   run: |
+      #     pipenv shell
       #     sceptre launch prod --yes
--- a/.github/workflows/aws-deploy.yaml
+++ b/.github/workflows/aws-deploy.yaml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
-      - uses: pre-commit/action@v2
+      - uses: pre-commit/action@v2.0
 
   sceptre-validate:
     name: Validate CloudFormation templates using sceptre
--- a/.github/workflows/aws-deploy.yaml
+++ b/.github/workflows/aws-deploy.yaml
@@ -26,4 +26,5 @@ jobs:
       - name: Deploy 'prod' templates
         uses: Rurquhart/sceptre-action@v1.0
         with:
-          sceptre_subcommand: launch --yes dev
+          sceptre_version: 2.4.0
+          sceptre_subcommand: launch --yes prod
--- a/.github/workflows/aws-deploy.yaml
+++ b/.github/workflows/aws-deploy.yaml
@@ -2,10 +2,6 @@ name: aws-deploy
 
 on: push
 
-env:
-  AWS_ACCESS_KEY_ID: ${{ secrets.NEXTFLOW_PROD_CI_ACCESS_KEY }}
-  AWS_SECRET_ACCESS_KEY: ${{ secrets.NEXTFLOW_PROD_CI_SECRET_ACCESS_KEY }}
-
 jobs:
   pre-commit:
     name: Run pre-commit hooks against all files
@@ -23,6 +19,14 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
+      - name: Assume AWS Role
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.NEXTFLOW_PROD_CI_ACCESS_KEY }}
+          aws-secret-access-key: ${{ secrets.NEXTFLOW_PROD_CI_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+          role-to-assume: arn:aws:iam::728882028485:role/workflows-nextflow-ci-service-account-ServiceRole-CTLXKJIOCRO4
+          role-duration-seconds: 1200
       - name: Deploy 'prod' templates
         uses: Rurquhart/sceptre-action@v1.0
         with:
--- a/.github/workflows/aws-deploy.yaml
+++ b/.github/workflows/aws-deploy.yaml
@@ -1,54 +1,41 @@
-name: aws-deploy
+name: Validate and Deploy CloudFormation Templates
 
-on:
-  pull_request:
-    branches: '*'
-  push:
-    branches: [prod]
+on: push
+
+env:
+  AWS_ACCESS_KEY_ID: ${{ secrets.NEXTFLOW_PROD_CI_ACCESS_KEY }}
+  AWS_SECRET_ACCESS_KEY: ${{ secrets.NEXTFLOW_PROD_CI_SECRET_ACCESS_KEY }}
 
 jobs:
-
   pre-commit:
+    name: Run pre-commit hooks against all files
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
-      - uses: pre-commit/action@v2.0.0
+      - uses: pre-commit/action@v2
 
-  sceptre-deploy:
-    if: github.event_name == 'push'
+  sceptre-validate:
+    name: Validate CloudFormation templates using sceptre
     runs-on: ubuntu-latest
     needs: pre-commit
     steps:
-      - uses: actions/checkout@v2
-      - name: Set up Python 3.8
-        uses: actions/setup-python@v1
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Validate 'prod' templates
+        uses: Rurquhart/sceptre-action@v1
         with:
-          python-version: 3.8
-      - name: Cache Pipenv virtualenv
-        uses: actions/cache@v2
-        id: pipenv-cache
+          sceptre_subcommand: validate prod
+
+  sceptre-deploy:
+    name: Deploy CloudFormation templates using sceptre
+    runs-on: ubuntu-latest
+    needs: sceptre-validate
+    if: github.ref == 'refs/heads/prod'
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Deploy 'prod' templates
+        uses: Rurquhart/sceptre-action@v1
         with:
-          path: ~/.local/share/virtualenvs
-          key: ${{ runner.os }}-pipenv-v2-${{ hashFiles('**/Pipfile.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-pipenv-v2-
-      - name: Install Dependencies
-        uses: VaultVulp/action-pipenv@v2.0.1
-        if: steps.pipenv-cache.outputs.cache-hit != 'true'
-        env:
-          PIPENV_NOSPIN: 'true'
-          WORKON_HOME: ~/.local/share/virtualenvs
-          PIPENV_CACHE_DIR: ~/.local/share/pipcache
-        with:
-          command: install --dev
-      - name: Login Using CI Service User
-        uses: aws-actions/configure-aws-credentials@v1
-        with:
-          aws-access-key-id: ${{ secrets.NEXTFLOW_PROD_CI_ACCESS_KEY }}
-          aws-secret-access-key: ${{ secrets.NEXTFLOW_PROD_CI_SECRET_ACCESS_KEY }}
-          aws-region: us-east-1
-      - name: Deploy to AWS Account
-        uses: VaultVulp/action-pipenv@v2.0.1
-        with:
-          command: run sceptre launch prod --yes
+          sceptre_subcommand: launch --yes dev
--- a/.github/workflows/aws-deploy.yaml
+++ b/.github/workflows/aws-deploy.yaml
@@ -23,7 +23,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
       - name: Validate 'prod' templates
-        uses: Rurquhart/sceptre-action@v1
+        uses: Rurquhart/sceptre-action@v1.0
         with:
           sceptre_subcommand: validate prod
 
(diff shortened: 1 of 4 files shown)


BrunoGrandePhD:
I'll request a review from @zaro0508 once he accepts the invitation to join this GitHub organization, which I just issued.

BrunoGrandePhD:
I played with the base branch to force-update the diffs, which were stuck on comparing with the old `develop` branch (before I made it up-to-date with `main`).

.github/workflows/aws-deploy.yaml
```diff
@@ -0,0 +1,76 @@
...
+          aws-access-key-id: ${{ secrets.CI_SERVICE_USER_ACCESS_KEY }}
+          aws-secret-access-key: ${{ secrets.CI_SERVICE_USER_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+          role-to-assume: arn:aws:iam::035458030717:role/SceptreBuildAccessRole
+          role-duration-seconds: 1200
+      - name: Deploy 'dev' and 'prod' Stack Groups to Dev Account
+        run: |
+          pipenv shell
+          sceptre launch dev --yes
+          sceptre launch prod --yes
```
tthyer, line 58: Do these three commands  work? 
I expected instead only this line: `pipenv run sceptre launch dev --yes`

.github/workflows/aws-deploy.yaml
```diff
@@ -0,0 +1,76 @@
...
+        run: |
+          pip install pipenv
+      - name: Cache Pipenv virtualenv
+        uses: actions/cache@v2
+        id: pipenv-cache
+        with:
+          path: ~/.local/share/virtualenvs
+          key: ${{ runner.os }}-pipenv-v2-${{ hashFiles('**/Pipfile.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-pipenv-v2-
```
tthyer, line 38: This is interesting, nice improvement

.github/workflows/aws-deploy.yaml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Install Dependencies
+        if: steps.pipenv-cache.outputs.cache-hit != 'true'
+        env:
+          PIPENV_NOSPIN: 'true'
+          WORKON_HOME: ~/.local/share/virtualenvs
+          PIPENV_CACHE_DIR: ~/.local/share/pipcache
+        run: pipenv install --dev
+      - name: Assume AWS Role (Dev Account)
+        uses: aws-actions/configure-aws-credentials@v1
```
tthyer, line 47: I was expecting to see the conditional `if: github.ref == 'refs/heads/develop'` on this step and the following.

.github/workflows/aws-deploy.yaml
```diff
@@ -0,0 +1,76 @@
...
+        with:
+          aws-access-key-id: ${{ secrets.CI_SERVICE_USER_ACCESS_KEY }}
+          aws-secret-access-key: ${{ secrets.CI_SERVICE_USER_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+          role-to-assume: arn:aws:iam::035458030717:role/SceptreBuildAccessRole
+          role-duration-seconds: 1200
+      - name: Deploy 'dev' and 'prod' Stack Groups to Dev Account
```
tthyer, line 54: Why are you launching both dev and prod to the Dev account? Was expecting dev -> dev and prod -> prod.

tthyer (changes requested):

.github/workflows/aws-deploy.yaml
```diff
@@ -0,0 +1,76 @@
...
+    if: github.event_name == 'push'
+    runs-on: ubuntu-latest
+    needs: pre-commit
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.8
+      - name: Install pipenv
```
zaro0508, line 28: maybe you want to use the github actions instead? https://github.com/marketplace/actions/pipenv-for-github-actions

.github/workflows/aws-deploy.yaml
```diff
@@ -0,0 +1,76 @@
...
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.CI_SERVICE_USER_ACCESS_KEY }}
+          aws-secret-access-key: ${{ secrets.CI_SERVICE_USER_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+          role-to-assume: arn:aws:iam::035458030717:role/SceptreBuildAccessRole
+          role-duration-seconds: 1200
+      - name: Deploy 'dev' and 'prod' Stack Groups to Dev Account
+        run: |
+          pipenv shell
```
zaro0508, line 56: again, pipenv actions might be better?

.github/workflows/aws-deploy.yaml
```diff
@@ -0,0 +1,76 @@
+name: aws-deploy
+
+on:
+  pull_request:
+    branches: '*'
+  push:
+    branches: [main, develop]
```
zaro0508, line 7: deploying cloudformation from 2 branches may mess you up because one might clobber the other.

zaro0508 (changes requested):

.github/workflows/aws-deploy.yaml
BrunoGrandePhD: I can add the `if` condition as long as we can guarantee that no commits will ever find themselves on `main` without first being on `develop`. Otherwise, you might end up with stacks that haven't been deployed to the dev account. 

Omitting the `if` condition seemed safer because on pushes to `main`, sceptre should skip deployment if a stack in `dev` already exists in the dev account, but if it doesn't, then it would "fill in the gaps". 

Not sure if my safer approach is needed though if we prevent commits to main (or are just diligent with not committing `dev` stacks directly to `main`).

.github/workflows/aws-deploy.yaml
BrunoGrandePhD: I explained my thinking for this approach [here](https://github.com/Sage-Bionetworks-Workflows/aws-workflows-nextflow-infra/pull/1#issue-612835313), but I rather not cause any confusion. So, I'll change this to use the approach you described. If the need arises in the future, we can add a `common` stack group for resources we want deployed in both accounts.

.github/workflows/aws-deploy.yaml
BrunoGrandePhD: I'll address this by conditionally deploying the `dev` and `prod` stack groups if the triggering branch is `develop` and `main`, respectively (using something like `if: github.ref == 'refs/heads/develop'`). I'm just trying to stay [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

.github/workflows/aws-deploy.yaml
tthyer: @BrunoGrandePhD You're right -- there is an opportunity for user error with the model that was suggested to you. It is the same model that is used for Service Catalog. You can see more information about how that development process works [here](https://sagebionetworks.jira.com/wiki/spaces/IT/pages/1122336799/Developer+Bootstrap). The assumption is that one will create templates in config/dev for testing (with merges to develop) and once those are tested, then the user duplicates the templates in prod and merges to main. For reference, also see the [scipool-infra travis build](https://github.com/Sage-Bionetworks/scipool-infra/blob/develop/.travis.yml).

That method is necessary when testing something that has a UI, as Service Catalog does. It is really not necessary if you're only deploying templates. If you're doing that, then a model that follows the pattern of [scicomp-provisioner](https://github.com/Sage-Bionetworks/scicomp-provisioner/tree/master/config) is probably more sensible -- the one difference from scicomp-provisioner being that one would have one set of configs as it does, but would merge to develop branch to test, and merge to main to deploy to the PHI account.

If you want to create your own development model, I think that's ok, but please document the process in a CONTRIBUTING.md file -- we'll need one of those anyway.

.github/workflows/aws-deploy.yaml
BrunoGrandePhD: I'll give it a try! We'll have to confirm that it interacts well with the `aws-actions/configure-aws-credentials@v1` action.

.github/workflows/aws-deploy.yaml
BrunoGrandePhD: I implemented what you suggested in your other comment (dev -> dev, prod -> prod) and switched to the [GitHub Action](https://github.com/Sage-Bionetworks-Workflows/aws-workflows-nextflow-infra/pull/1#discussion_r612507300) that Khai suggested using `pipenv run`.

BrunoGrandePhD:
@zaro0508: Ready for a second round of review if you want.

.github/workflows/aws-deploy.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+      # Commented out until the AWS account becomes available. Track this PR:
+      # https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/135
+
+      # Deployment to Production AWS Account
+      # - name: Assume AWS Role (Production Account)
+      #   uses: aws-actions/configure-aws-credentials@v1
+      #   if: github.ref == 'refs/heads/main'
+      #   with:
+      #     aws-access-key-id: ${{ secrets.CI_SERVICE_USER_ACCESS_KEY }}
```
zaro0508, line 69: you'll probably need to setup a separate set of ENV Vars for prod account

.github/workflows/aws-deploy.yaml
```diff
@@ -0,0 +1,78 @@
...
+        with:
+          command: run sceptre launch dev --yes
+
+      # Commented out until the AWS account becomes available. Track this PR:
+      # https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/135
+
+      # Deployment to Production AWS Account
+      # - name: Assume AWS Role (Production Account)
+      #   uses: aws-actions/configure-aws-credentials@v1
+      #   if: github.ref == 'refs/heads/main'
```
zaro0508, line 67: consider replacing `main` with `prod` branch?

zaro0508 (approved):

.github/workflows/aws-deploy.yaml
BrunoGrandePhD: I took this pattern from the `organizations-infra` [repo](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/.github/workflows/aws-deploy.yaml), where the same service user assumes different roles for deploying to each account. 

@zaro0508: Is there an advantage with using one service user with different roles per account vs different service users per account?

.github/workflows/aws-deploy.yaml
BrunoGrandePhD: Good idea!

BrunoGrandePhD:
For the time being, I propose that we set up this repository to only deploy to the production AWS account. We can reserve the development account for testing templates using manual sceptre deployment. 

I've simplified the CI workflow in this PR accordingly. I'm open to suggestions on how to further improve this workflow. 

/cc @zaro0508 @tthyer

.github/workflows/aws-deploy.yaml
```diff
@@ -0,0 +1,54 @@
...
+    if: github.event_name == 'push'
+    runs-on: ubuntu-latest
+    needs: pre-commit
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.8
+      - name: Cache Pipenv virtualenv
```
zaro0508, line 28: hmm, wouldn't it be better to decouple this cache from this job?  maybe create the cache in separate job and have downstream jobs depend on it?

.github/workflows/aws-deploy.yaml
```diff
@@ -0,0 +1,54 @@
...
+      - name: Login Using CI Service User
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.NEXTFLOW_PROD_CI_ACCESS_KEY }}
+          aws-secret-access-key: ${{ secrets.NEXTFLOW_PROD_CI_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+      - name: Deploy to AWS Account
+        uses: VaultVulp/action-pipenv@v2.0.1
+        with:
+          command: run sceptre launch prod --yes
```
zaro0508, line 54: ahh, cool. i just noticed this.. https://github.com/marketplace/actions/sceptre-github-action

zaro0508 (changes requested):

.github/workflows/aws-deploy.yaml
tthyer: That's cool too -- going to start using that one and the pipenv one

tthyer (approved):
Other than perhaps also using the newly discovered sceptre action, this LGTM

.github/workflows/aws-deploy.yaml
BrunoGrandePhD: I think the Sceptre GitHub action eliminates the need for the caching step since the action uses a Docker container with Sceptre preinstalled.

.github/workflows/aws-deploy.yaml
BrunoGrandePhD: Another good find! This actually streamlines the CI workflow considerably.

.github/workflows/aws-deploy.yaml
```diff
@@ -0,0 +1,41 @@
...
+jobs:
+  pre-commit:
+    name: Run pre-commit hooks against all files
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - uses: pre-commit/action@v2.0.2
+
+  sceptre-validate:
```
zaro0508, line 18: i believe `sceptre validate` just calls `aws cloudformation validate` which requires access to AWS.  That's gonna fail here since this job won't have access to AWS.  This will work if you run as post merge step though (i.e. add `if: github.ref == 'refs/heads/prod'`).  

With that being said I think you don't really need validate at all since it's basically doing the same thing as cfn-lint which is already being run with pre-commit.

.github/workflows/aws-deploy.yaml
```diff
@@ -0,0 +1,41 @@
+name: aws-deploy
+
+on: push
+
+env:
+  AWS_ACCESS_KEY_ID: ${{ secrets.NEXTFLOW_PROD_CI_ACCESS_KEY }}
```
zaro0508, line 6: you'll need to setup a service user for this, you can use this template to setup a user/role combination for this.  https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/IAM/service-account.yaml

hmm, we prefer the CI to assume a role to do it's thing and that doesn't look like something you can do with the sceptre GH action.  i think we have 3 options:
1. create ~/.aws/config and ~/.aws/credentials files and set a AWS_PROFILE env var.
2. use the GH assume role action, not sure if that's gonna work with sceptre GH action though.
3. add the feature to sceptre GH action to assume a role.

zaro0508 (changes requested):

.github/workflows/aws-deploy.yaml
BrunoGrandePhD: I also found out that `sceptre validate` doesn't handle hooks, so I removed this step from the CI workflow.

.github/workflows/aws-deploy.yaml
BrunoGrandePhD: @zaro0508: Do you think specifying the [role ARN](https://sceptre.cloudreach.com/2.4.0/docs/stack_config.html#role-arn) in the sceptre config would help circumvent the lack of support for roles in the sceptre action?

zaro0508: ahh yes, good thinking @BrunoGrandePhD, i think that should work.  the role arn doesn't need to be a secret

.github/workflows/aws-deploy.yaml
```diff
@@ -0,0 +1,29 @@
...
+    runs-on: ubuntu-latest
+    needs: pre-commit
+    if: github.ref == 'refs/heads/prod'
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Deploy 'prod' templates
+        uses: Rurquhart/sceptre-action@v1.0
+        with:
+          sceptre_subcommand: launch --yes dev
```
tthyer, line 29: think this should be `launch --yes prod`

tthyer (changes requested):

.github/workflows/aws-deploy.yaml
zaro0508: also it looks like the default is `sceptre_version` is 2.3.0, https://github.com/Rurquhart/sceptre-action/blob/master/action.yml#L14

however the latest version is 2.4.0, https://github.com/Sceptre/sceptre/tags

we should be using the latest version.

zaro0508 (changes requested):

.github/workflows/aws-deploy.yaml
BrunoGrandePhD: Thanks to both of you!

.github/workflows/aws-deploy.yaml
BrunoGrandePhD: @zaro0508: After taking a look at the `aws-actions/configure-aws-credentials@v1` [action](https://github.com/aws-actions/configure-aws-credentials/blob/master/index.js#L114-L121) and the sceptre [codebase](https://github.com/Sceptre/sceptre/blob/v2.4.0/sceptre/connection_manager.py#L139-L141), I think the `aws-actions/configure-aws-credentials@v1` action will work. 

I'm holding off on using the `role_arn` config value because I prefer keeping everything related to the CI within that YAML file, if possible.

BrunoGrandePhD:
Now that Sage-Bionetworks-IT/organizations-infra#141 has been merged, I've updated the service user credentials in the organization secrets accordingly. I'm also using the `aws-actions/configure-aws-credentials@v1` action here to assume the service role that was created. From [what I can tell](https://github.com/Sage-Bionetworks-Workflows/aws-workflows-nextflow-infra/pull/1#discussion_r613634624), this action should be compatible with the sceptre action. 🤞 

@zaro0508 @tthyer: Feel free to take one last review.

tthyer (approved):
looks good!

zaro0508 (approved):
Yep, looks like the best option yet.  very nice and neat workflow.

zaro0508:
@thomasyu888 should really take a look at this PR

tthyer:
+1 there are several new actions that will be good to apply in our templates and ci examples, too

PR #2, opened 2021-04-13
Create S3 bucket for storing CloudFormation templates

BrunoGrandePhD:
For now, the only use case is for storing the templates from the [aws-genomics-workflows-library](https://github.com/Sage-Bionetworks-Workflows/aws-genomics-workflows-library) repository. 

@zaro0508: Is there a reason why you deploy templates to an S3 bucket instead of downloading from GitHub directly in the sceptre pre-hooks? If you think the S3 deployment is essential, then your review of this PR would be appreciated.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -1,2 +1,3 @@
-project_code: aws-nextflow-workflows
+project_code: aws-workflows-nextflow-infra
 region: us-east-1
+admincentral_cf_bucket: bootstrap-awss3cloudformationbucket-19qromfd235z9
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -1,3 +1,4 @@
 project_code: aws-workflows-nextflow-infra
-region: us-east-1
-admincentral_cf_bucket: bootstrap-awss3cloudformationbucket-19qromfd235z9
+profile: {{ var.profile | default("default") }}
+region: {{ var.region | default("us-east-1") }}
+aws_infra_templates_root_url: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.2.3/templates
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -1,4 +1,4 @@
 project_code: aws-workflows-nextflow-infra
 profile: {{ var.profile | default("default") }}
 region: {{ var.region | default("us-east-1") }}
-aws_infra_templates_root_url: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.2.3/templates
+aws_infra_templates_root_url: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra
--- a/config/prod/s3-cfn-templates.yaml
+++ b/config/prod/s3-cfn-templates.yaml
@@ -1,0 +1,9 @@
+template_path: "remote/S3/public-bucket.yaml"
+stack_name: "cloudformation-templates"
+stack_tags:
+  Department: "CompOnc"
+  Project: "imCORE"
+  OwnerEmail: "bruno.grande@sagebase.org"
+hooks:
+  before_launch:
+    - !cmd "curl https://{{stack_group_config.admincentral_cf_bucket}}.s3.us-east-1.amazonaws.com/aws-infra/master/S3/public-bucket.yaml --create-dirs -o templates/remote/S3/public-bucket.yaml"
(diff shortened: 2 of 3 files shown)


zaro0508:
> @zaro0508: Is there a reason why you deploy templates to an S3 bucket instead of downloading from GitHub directly in the sceptre pre-hooks? If you think the S3 deployment is essential, then your review of this PR would be appreciated.

Nope, you don't need to download from S3.  You can download directly from raw github file references when using sceptre.  Cases where you may want to put templates in S3 is when you use a cloudformation type that only supports S3 references.  For example lambda code, includes of template snippets, and few other things that i can't recall ATM.  

The question here is whether you want your own cloudformation repo or do you just want to use the already existing Sage CFN repo? https://bootstrap-awss3cloudformationbucket-19qromfd235z9.s3.amazonaws.com/index.html

tthyer (changes requested):
@BrunoGrandePhD I think I prefer just using a raw github url

BrunoGrandePhD:
I've updated this PR to use the raw GitHub URL pinned to a version of the repository. This PR will also serve to test the `aws-deploy` CI workflow that was just merged. At the moment, the CI is failing because sceptre returns exit code 1 as long as the `prod` stack group is empty. 

/cc @tthyer @zaro0508

BrunoGrandePhD:
> The question here is whether you want your own cloudformation repo or do you just want to use the already existing Sage CFN repo? https://bootstrap-awss3cloudformationbucket-19qromfd235z9.s3.amazonaws.com/index.html

@zaro0508: I think our own bucket would be ideal because I want to leverage the existing [Travis CI deployment workflow](https://github.com/Sage-Bionetworks-Workflows/aws-genomics-workflows-library/blob/master/.travis.yml) in the `aws-genomics-workflows-library` repository and I would like to keep that separate from the existing CFN S3 bucket.

config/config.yaml
```diff
@@ -1,2 +1,4 @@
-project_code: aws-nextflow-workflows
-region: us-east-1
+project_code: aws-workflows-nextflow-infra
+profile: {{ var.profile | default("default") }}
+region: {{ var.region | default("us-east-1") }}
+aws_infra_templates_root_url: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.2.3/templates
```
zaro0508, line 4: I suggest the root be `https://raw.githubusercontent.com/Sage-Bionetworks` because there are things outside of aws-infra you might be interested, like lambdas published from Sage-Bionetworks-IT org.  Also you may want to reference different versions of  the CFN templates in each one of your sceptre templates.

config/prod/s3-cfn-templates.yaml
```diff
@@ -0,0 +1,9 @@
+template_path: "remote/S3/public-bucket.yaml"
+stack_name: "cloudformation-templates"
+stack_tags:
+  Department: "CompOnc"
+  Project: "imCORE"
+  OwnerEmail: "bruno.grande@sagebase.org"
+hooks:
+  before_launch:
+    - !cmd "curl {{stack_group_config.aws_infra_templates_root_url}}/S3/public-bucket.yaml --create-dirs -o templates/remote/S3/public-bucket.yaml"
```
zaro0508, line 9: please use `wget` instead of `curl`.  it's better because it has retries built in. here's an example, https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/sceptre/scicomp/config/prod/rotate-credentials.yaml#L12

zaro0508 (changes requested):

config/config.yaml
zaro0508: ohh wait, sorry.  ignore this comment.  i got mixed up with github reference and S3 bucket reference.  this is ok.

BrunoGrandePhD: Got it!

zaro0508: sorry, the version thing i said in the first comment if valid, so i think the root should be.. `https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra`

BrunoGrandePhD:
@zaro0508: I switched to `wget`. Thanks for the tip!

zaro0508 (approved):

BrunoGrandePhD:
@tthyer: I switched to the GitHub URL. I'll still wait for your approval before merging.

tthyer (approved):

PR #5, opened 2021-04-15
Add VPC

tthyer:
Add a VPC for NextFlow.

### DIFF
--- a/config/prod/nextflow-vpc.yaml
+++ b/config/prod/nextflow-vpc.yaml
@@ -1,0 +1,8 @@
+template_path: remote/vpc.yaml
+stack_name: nextflow-vpc
+parameters:
+  VpcName: nextflow-vpc
+  VpcSubnetPrefix: "10.41"
+hooks:
+  before_launch:
+    - !cmd "wget {{stack_group_config.aws_infra_templates_root_url}}/v0.2.3/templates/vpc.yaml -N -P templates/remote"


BrunoGrandePhD (approved):
🚀 

For my own curiosity, we're not bothering with tagging the VPC with `Department` and `Project` because it doesn't cost anything, right?

tthyer:
@BrunoGrandePhD Deliberately not tagging because these have default values that seem more appropriate (i.e., IT infrastructure) rather than to a Science department and individual. See the template starting at [line 27](https://github.com/Sage-Bionetworks/aws-infra/blob/v0.2.3/templates/vpc.yaml#L27).

config/prod/nextflow-vpc.yaml
```diff
@@ -0,0 +1,8 @@
+template_path: remote/vpc.yaml
+stack_name: nextflow-vpc
+parameters:
+  VpcName: nextflow-vpc
+  VpcSubnetPrefix: "10.41"
```
zaro0508, line 5: Here is our list of VPCs https://sagebionetworks.jira.com/wiki/spaces/IT/pages/352976898/Sage+VPN.  

You'll just need to pick a value that isn't on the list, so how about  `172.21`?   When you decide on a value please add it to the wiki.

zaro0508 (changes requested):

zaro0508 (changes requested):
Can we hold off on this?  IT needs to think about this a little more.  We think using a vpc.yaml for this is not appropriate and are discussing alternatives.

tthyer:
Closing this pull request -- will open a new one when a new [medium-sized VPC template](https://github.com/Sage-Bionetworks/aws-infra/pull/293) is available.

PR #7, opened 2021-04-30
CICD refactor

tthyer:
We agreed to change the pattern for how we will organize and deploy configurations into three categories: common, develop, and prod(uction).

- Since there were already candidate configurations for common in the IAM policies and user, that division was created.
- CICD was updated to deploy multiple configurations on merges to "main" branch.
- Unfortunately, the sceptre action appears to take only one subcommand at at time -- a disadvantage to using this action over a more flexible script.


.github/workflows/aws-deploy.yaml
```diff
@@ -27,7 +27,17 @@ jobs:
...
+      - name: Deploy common configuration
+        uses: Sage-Bionetworks/sceptre-action@v1.1
+        with:
+          sceptre_version: 2.4.0
+          sceptre_subcommand: launch --yes common
+      - name: Deploy develop configuration
+        uses: Sage-Bionetworks/sceptre-action@v1.1
+        with:
+          sceptre_version: 2.4.0
+          sceptre_subcommand: launch --yes develop
```
zaro0508, line 39: i believe this will fail if you don't have a `config/develop` folder and files to deploy.

zaro0508 (changes requested):

.github/workflows/aws-deploy.yaml
BrunoGrandePhD: @tthyer: You can leave this block commented out. I have a PR in mind that I can open once this is merged that would only deploy to the dev account.

BrunoGrandePhD: @zaro0508: I just realized that Tess renamed one of the `prod` files to `config/develop/s3-cfn-templates.yaml`, so we should be good.

BrunoGrandePhD: However, the `prod` folder is [empty](https://github.com/Sage-Bionetworks-Workflows/aws-workflows-nextflow-infra/tree/tthyer/cicd-refactor/config) now, so that might fail.

BrunoGrandePhD (approved):
Aside from the empty `prod` folder, all looks good to me!  🚀

.github/workflows/aws-deploy.yaml
zaro0508: yup, and you don't need that empty `prod/config.yaml` file either.

zaro0508 (approved):

PR #12, opened 2021-05-07
Provision service user for accessing S3 data

BrunoGrandePhD:
I'm using this service user for accessing data files stored in S3. I'm provisioning it in the dev account to avoid granting the user full S3 access in the prod account. I prefer provisioning the user here as oppose to sandbox just to keep Nextflow infrastructure together.


tthyer (approved):
@BrunoGrandePhD Approved -- is this following the model where we create separate service users per actual user to isolate their access to Synapse config files stored in S3? Or other?

BrunoGrandePhD:
I intend to use bucket policies to grant this IAM user with access to S3 buckets, _i.e._ some within the same account (_i.e._ nextflow-dev) and some in others accounts (_e.g._ sandbox). I don't need the user to access all of the buckets in the nextflow-dev account. 

I expect `AmazonS3FullAccess` to grant this user with access to all buckets within nextflow-dev. Would an inline policy that doesn't list any resources work if access is granted via bucket policies (see example below)? Or do I need to list the individual buckets (even the cross-account ones)? 

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": "s3:*",
            "Resource": []
        }
    ]
}
```

config/develop/bgrande-nextflow-service-user.yaml
```diff
@@ -0,0 +1,12 @@
+# Provision a service account for AWS API access keys
+template_path: "remote/service-user.yaml"
+stack_name: "bgrande-nextflow-service-user"
+stack_tags:
+  Department: "CompOnc"
+  Project: "imCORE"
+  OwnerEmail: "bruno.grande@sagebase.org"
+parameters:
+  ManagedPolicyArns: ["arn:aws:iam::aws:policy/AmazonS3FullAccess"]
```
zaro0508, line 9: you can be more specific about which bucket(s) this service user has access to by providing a custom managed or inline policy specifying the specific buckets.  here's an example.. https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/sceptre/sageit/config/prod/gha-svc-account.yaml

BrunoGrandePhD: Thanks for the example!

zaro0508 (approved):

tthyer (approved):

PR #15, opened 2021-06-11
Add an AWS Aurora cluster required for NextFlow Tower app

tthyer:
For [WORKFLOWS-10](https://sagebionetworks.jira.com/browse/WORKFLOWS-10), this creates a serverless Aurora cluster in the nextflow VPC using two private subnets. The Data API is enabled, which allows us to use awscli to query the database, after authenticating to a role with permissions. For example,

```
aws  rds-data execute-statement \
  --resource-arn $AURORA_CLUSTER_ARN \
  --schema mysql \
  --secret-arn $AURORA_SECRET_ARN \
  --sql 'SHOW DATABASES;'
```

I have tested spinning this stack up with and without a snapshot in the nextflow-dev account.

Related planned work:
- Bootstrapping, if we intend to do that through CFN, will be set up in [WORKFLOWS-12](https://sagebionetworks.jira.com/browse/WORKFLOWS-12).
- A logging system will be set up in [WORKFLOWS-17](https://sagebionetworks.jira.com/browse/WORKFLOWS-17).
- Secret rotation will be set up in [WORKFLOWS-16](https://sagebionetworks.jira.com/browse/WORKFLOWS-16).

### DIFF
--- a/config/common/nextflow-aurora-mysql.yaml
+++ b/config/common/nextflow-aurora-mysql.yaml
@@ -1,0 +1,7 @@
+template_path: nextflow-aurora-mysql.yaml
+stack_name: nextflow-aurora-mysql
+parameters:
+  VpcID: !stack_output_external nextflow-vpc::VPCId
+  SubnetIDs:
+    - !stack_output_external nextflow-vpc::PrivateSubnet1
+    - !stack_output_external nextflow-vpc::PrivateSubnet2
--- a/config/develop/nextflow-aurora-mysql.yaml
+++ b/config/develop/nextflow-aurora-mysql.yaml
@@ -1,5 +1,7 @@
 template_path: nextflow-aurora-mysql.yaml
 stack_name: nextflow-aurora-mysql
+dependencies:
+  - develop/nextflow-vpc.yaml
 parameters:
   VpcID: !stack_output_external nextflow-vpc::VPCId
   SubnetIDs:
--- a/config/develop/nextflow-aurora-mysql.yaml
+++ b/config/develop/nextflow-aurora-mysql.yaml
@@ -7,3 +7,7 @@ parameters:
   SubnetIDs:
     - !stack_output_external nextflow-vpc::PrivateSubnet1
     - !stack_output_external nextflow-vpc::PrivateSubnet2
+stack_tags:
+  Department: IBC
+  Project: Infrastructure
+  OwnerEmail: nextflow-admins@sagebase.org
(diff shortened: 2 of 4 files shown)


thomasyu888 (approved):
Looks good to me.  I did see a set of instructions:

> NOTE: Ensure the RDS database and ECS Cluster on the same VPC/SG with the 3306 port inbound open

Does this have to be explicitly added anywhere, or are all inbound ports open behind this VPC?

config/common/nextflow-aurora-mysql.yaml
```diff
@@ -0,0 +1,7 @@
+template_path: nextflow-aurora-mysql.yaml
+stack_name: nextflow-aurora-mysql
+parameters:
+  VpcID: !stack_output_external nextflow-vpc::VPCId
```
zaro0508, line 4: why is this in `config/common`?  don't you want to create the DB in both dev and prod accounts?  

This depends on the VPC in nextflow-vpc.yaml file so please add a `dependencies` section for it.   Also `stack_tags` section would be good as well.

zaro0508 (changes requested):

tthyer:
Regarding:
> Looks good to me. I did see a set of instructions:
> 
> > NOTE: Ensure the RDS database and ECS Cluster on the same VPC/SG with the 3306 port inbound open
> 
> Does this have to be explicitly added anywhere, or are all inbound ports open behind this VPC?

@thomasyu888
The instructions for setting up the entire system are vague and only part of it is templatized, so there will be some back and forth with them. Rob said they want to go in that direction, so maybe there will be an update before we start setting everything up; I'm hoping I hear more about that this week before I have to go further down this path.

So for example one part is to create an EC2 instance, presumably to run the tower software, and that is what needs to make the connection. I would prefer to create that stack first with its own security group, and then create this Aurora stack, adding to that a SecurityGroupIngress on the ClusterSecurityGroup.

config/common/nextflow-aurora-mysql.yaml
tthyer: @zaro0508 You're right, although this works nicely in development due to the way the CICD deploys I'll have to duplicate this. Bummer.

Will add dependencies section. Please indicate which stack_tags you thing are appropriate. This will be shared by many projects.

zaro0508 (approved):

tthyer:
@zaro0508 stack tags added, merging now

PR #17, opened 2021-06-26
Use `-O` option for storing `wget` output to avoid caching

BrunoGrandePhD:
Prompted by Sage-Bionetworks/scicomp-provisioner#487.

### DIFF
--- a/.github/workflows/aws-deploy.yaml
+++ b/.github/workflows/aws-deploy.yaml
@@ -19,6 +19,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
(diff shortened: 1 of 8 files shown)


zaro0508 (changes requested):

tthyer (approved):

BrunoGrandePhD:
@tthyer @zaro0508: Can either of you confirm that a folder created during a GH CI workflow persists in subsequent steps? I'm 95% sure that the file system persists between steps but not between jobs, but I still would like to confirm.

zaro0508 (approved):
yes, file system persists between steps but not jobs.

PR #19, opened 2021-07-01
WORKFLOWS-25: support sceptre resolvers in build

tthyer:
The current github actions workflow uses a sceptre action which does not support custom resolvers. This is limits how we have come to use sceptre -- for example, we won't be able to use the ever-popular `sceptre-ssm-resolver`. I have use cases for using `sceptre-cmd-resolver`, so I have refactored the workflow to use the pipenv environment instead and forgo the sceptre action.

I also want to make more of a case for using pipenv in this way. It makes sense that you set up your dependencies file -- whether that is `requirements.txt` or `Pipfile` -- that can be used in any environment. In other words, it creates an environment that I can use to build whether I am developing or the build action is running on github. When we use the sceptre custom action, we're effectively departing from the simplicity of a single build environment.

This solves blocking issue [WORKFLOWS-25](https://sagebionetworks.jira.com/browse/WORKFLOWS-25).

### DIFF
--- a/.github/workflows/aws-deploy.yaml
+++ b/.github/workflows/aws-deploy.yaml
@@ -17,11 +17,32 @@ jobs:
     needs: pre-commit
     if: github.ref == 'refs/heads/main'
     steps:
+
       - name: Checkout repository
         uses: actions/checkout@v2
+
       - name: Create directory for remote sceptre templates
-        run:  mkdir -p templates/remote/
-      - name: Assume AWS Role
+        run: mkdir -p templates/remote/
+
+      - name: Set up python 3.9
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+
+      - name: Set up dependency caching
+        uses: actions/cache@v2
+        with:
+          path: ~/.local/share/virtualenvs
+          key: ${{ runner.os }}-pipenv-${{ hashFiles('Pipfile.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-pipenv-
+
+      - name: Install dependencies
+        run: |
+          python -m pip install pipenv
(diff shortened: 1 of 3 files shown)


tthyer:
@daisyhan97 Here's another PR that may interest you

Pipfile
```diff
@@ -4,10 +4,11 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-sceptre = "2.4.0"
+sceptre = "2.5.0"
```
zaro0508, line 7: shouldn't these be in `dev-packages`?

.github/workflows/aws-deploy.yaml
```diff
@@ -17,43 +17,57 @@ jobs:
...
+        with:
+          path: ~/.local/share/virtualenvs
+          key: ${{ runner.os }}-pipenv-${{ hashFiles('Pipfile.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-pipenv-
+
+      - name: Install dependencies
+        run: |
+          python -m pip install pipenv
+          pipenv install
```
zaro0508, line 43: shouldn't this be `pipenv install --dev`?

zaro0508 (approved):
looks fine in general, just had a few questions about pipenv packaing

BrunoGrandePhD (approved):
🚀

Pipfile
tthyer: It could be in dev, but then the only contents of the Pipfile would be in dev.

.github/workflows/aws-deploy.yaml
tthyer: see above

PR #35, opened 2021-07-21
Add stack that produces SMTP credentials

tthyer:
[WORKFLOWS-11](https://sagebionetworks.jira.com/browse/WORKFLOWS-11): NextFlow needs SMTP credentials set up so that Tower can send email.

This is adapated from [tsamuel33/aws-smtp-credentials](https://github.com/tsamuel33/aws-smtp-credentials).

### DIFF
--- a/templates/smtp-credentials.yaml
+++ b/templates/smtp-credentials.yaml
@@ -24,6 +24,13 @@ Parameters:
       URL of S3 bucket where templates used for nested stacks are deployed
     ConstraintDescription: Must be a valid S3 HTTP URL
 
+Mappings:
+  Accounts:
+    "035458030717":
+      AdminRole: "arn:aws:iam::035458030717:role/aws-reserved/sso.amazonaws.com/AWSReservedSSO_Administrator_580e9f32ac55c4e7"
+    "728882028485":
+      AdminRole: "arn:aws:iam::728882028485:role/aws-reserved/sso.amazonaws.com/AWSReservedSSO_Administrator_3a9d61d0884be260"
+
 Resources:
 
   SmtpUserGroup:
@@ -117,11 +124,13 @@ Resources:
           - - !Sub 'arn:aws:iam::${AWS::AccountId}:root'
             - !ImportValue
               'Fn::Sub': '${AWS::Region}-workflows-nextflow-ci-service-account-ServiceUserArn'
+            - !FindInMap [Accounts, !Ref AWS::AccountId, AdminRole]
         UserPrincipalArns: !Join
           - ','
           - - !ImportValue
               'Fn::Sub': '${AWS::Region}-workflows-nextflow-ci-service-account-ServiceUserArn'
             - !GetAtt SmtpLambdaExecutionRole.Arn
+            - !FindInMap [Accounts, !Ref AWS::AccountId, AdminRole]
 
   SmtpLambdaLogGroup:
     Type: AWS::Logs::LogGroup
--- a/templates/smtp-credentials.yaml
+++ b/templates/smtp-credentials.yaml
@@ -1,0 +1,272 @@
+# Derived from a template created by Travis Samuel
+# MIT License https://github.com/tsamuel33/aws-smtp-credentials/blob/main/LICENSE
+# Template repository https://github.com/tsamuel33/aws-smtp-credentials
+
+AWSTemplateFormatVersion: 2010-09-09
+Description: >-
+  This template creates a custom SMTP resource that stores the SMTP
+  credentials in Parameter Store under the names "smtp-username" and
+  "smtp-password".
+
+Parameters:
+
+  IamAccessKeyVersion:
+    Type: Number
+    Default: 1
+    Description: >-
+      Version number of the AWS access keys.
+      Increment this number to rotate the keys.
+    MinValue: 1
+
+  TemplateRootUrl:
+    Type: String
+    Description: >-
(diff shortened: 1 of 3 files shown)


thomasyu888 (approved):

templates/smtp-credentials.yaml
```diff
@@ -0,0 +1,272 @@
...
+    Type: AWS::Lambda::Function
+    DependsOn: SmtpLambdaLogGroup
+    Properties:
+      Description: Generates SMTP credentials and stores in Parameter Store
+      FunctionName: SMTPCredentialsLambdaFunction
+      Handler: index.lambda_handler
+      MemorySize: 128
+      Environment:
+        Variables:
+          KMS_KEY_ARN: !Sub 'arn:aws:kms:${AWS::Region}:${AWS::AccountId}:key/${SmtpKeyStack.Outputs.Key}'
```
zaro0508, line 141: this is a dependency on `SmtpKeyStack`.  i would add `SmtpKeyStack` to `DependsOn` in this resource.

templates/smtp-credentials.yaml
```diff
@@ -0,0 +1,272 @@
...
+            Statement:
+                - Effect: Allow
+                  Action:
+                    - logs:CreateLogStream
+                    - logs:PutLogEvents
+                  Resource:
+                    - !Sub "arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/lambda/SMTPCredentialsLambdaFunction"
+                    - !Sub "arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/lambda/SMTPCredentialsLambdaFunction:log-stream:*"
+                - Effect: Allow
+                  Action:
```
zaro0508, line 101: it looks like the calculate_key() function in the lambda code is doing some decrypting to get the pwd, so don't you want the policy from the original code?
```
                - Effect: Allow
                  Action:
                    - kms:Decrypt
                    - kms:DescribeKey
                    - kms:CreateGrant
                    - kms:Encrypt
                  Resource:
                    - !Ref KmsKeyArn
```

templates/smtp-credentials.yaml
```diff
@@ -0,0 +1,272 @@
...
+    Properties:
+      TemplateURL: !Sub ${TemplateRootUrl}/aws-infra/v0.2.19/KMS/kms-key.yaml
+      TimeoutInMinutes: 5
+      Parameters:
+        AliasName: 'alias/smtp-credentials'
+        AdminPrincipalArns: !Join
+          - ','
+          - - !Sub 'arn:aws:iam::${AWS::AccountId}:root'
+            - !ImportValue
+              'Fn::Sub': '${AWS::Region}-workflows-nextflow-ci-service-account-ServiceUserArn'
```
zaro0508, line 119: you may want to give some humans access to this key?  otherwise you may need need to use the CI service account to debug problems with this.

zaro0508 (changes requested):

templates/smtp-credentials.yaml
tthyer: yes, was just dragging my feet on that hoping to be able to `!ImportValue`

templates/smtp-credentials.yaml
tthyer: Isn't that implicit as usual, since I'm using `!Sub`?

templates/smtp-credentials.yaml
tthyer: No, I added SmtpLambdaExecutionRole as a prinicipal on the SmtpKeyStack. The way he wrote it actually doesn't work -- there was a change in 2019.

zaro0508:
Sorry this slipped my mind while reviewing but here's a more general question.. is there a reason why you are doing this with a lambda & param store instead of just saving the SMTP credentials to the secrets manager?  CFN  for the secrets manager [natively supports generating secrets](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-secretsmanager-secret.html#cfn-secretsmanager-secret-generatesecretstring) which would make it much, much simpler.   Take a look at the `GenerateSecretString` property

tthyer:
@zaro0508, I am not doing this from scratch. I am using an existing stack that you pointed me to and making only the fixes and alterations necessary at this time. I don't want to rewrite it at this time.

tthyer:
@zaro0508 Can you approve this PR or continue conversation? I'm not planning a refactor to use secretsmanager.

templates/smtp-credentials.yaml
zaro0508: I don't think `!Sub` makes it implicit, i think it depends on the property that you pass to.  these are arbitrary env vars so CFN may not enforce it as a dependency which may cause this to fail if timing is off.  my recommendation is to add to `DependsOn`

zaro0508 (commented):
My recommendation is to use the secrets manager, not only because it will makes this much simpler and more maintainable going forward but also that there are also a host of benefits to be gained by using the secrets manager over the parameter store.  This is my recommendation however you may do as you wish.

templates/smtp-credentials.yaml
tthyer: I haven't seen this be a problem and from my reading of the docs this should work the same as any other use of implicit depends through Ref (which Sub is using) or GetAtt

PR #36, opened 2021-08-09
WORKFLOWS-20: Custom S3 Bucket Template

daisyhan97:
First draft of S3 bucket template.
Private by default, grants access to specified users using ARNs

### DIFF
--- a/templates/nextflow-s3-template.yaml
+++ b/templates/nextflow-s3-template.yaml
@@ -3,21 +3,6 @@ Description: Private S3 Bucket for Nextflow
 
 # Parameters
 Parameters:
-  Department:
-    Description: 'The department for this resource'
-    Type: String
-    AllowedPattern: '^\S*$'
-    ConstraintDescription: 'Must be string with no spaces'
-  Project:
-    Description: 'The name of the project that this resource is used for'
-    Type: String
-    AllowedPattern: '^\S*$'
-    ConstraintDescription: 'Must be string with no spaces'
-  OwnerEmail:
-    Description: 'Email address of the owner of this resource'
-    Type: String
-    AllowedPattern: '^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$'
-    ConstraintDescription: 'Must be an acceptable email address syntax(i.e. joe.smith@sagebase.org)'
   BucketName:
     Type: String
     Description: (Optional) Name of the created bucket.
--- a/templates/nextflow-s3-template.yaml
+++ b/templates/nextflow-s3-template.yaml
@@ -89,10 +89,14 @@ Resources:
             # gives grantees access to the bucket
             Effect: Allow
             Principal:
-              AWS: !Join
+              AWS: !Split
                 - ","
-                - - Fn::ImportValue: !Sub "${AWS::Region}-nextflow-forge-service-user-ServiceUserArn"
-                  - !Join [",", !Ref BucketAccess]
+                - !Join
+                    - ","
+                    - - Fn::ImportValue: !Sub "${AWS::Region}-nextflow-forge-service-user-ServiceUserArn"
+                      - !Join
+                          - ","
+                          - !Ref BucketAccess
             Action:
               - "s3:ListBucket"
               - "s3:ListBucket*"
@@ -105,10 +109,14 @@ Resources:
             # give grantees read access to objects
             Effect: Allow
             Principal:
-              AWS: !Join
+              AWS: !Split
                 - ","
-                - - Fn::ImportValue: !Sub "${AWS::Region}-nextflow-forge-service-user-ServiceUserArn"
-                  - !Join [",", !Ref ReadAccess]
+                - !Join
+                    - ","
+                    - - Fn::ImportValue: !Sub "${AWS::Region}-nextflow-forge-service-user-ServiceUserArn"
+                      - !Join
+                          - ","
+                          - !Ref ReadAccess
             Action:
               - "s3:GetObject"
               - "s3:GetObjectAcl"
@@ -121,10 +129,14 @@ Resources:
           - Sid: WriteAccess
             Effect: Allow
             Principal:
-              AWS: !Join
+              AWS: !Split
                 - ","
-                - - Fn::ImportValue: !Sub "${AWS::Region}-nextflow-forge-service-user-ServiceUserArn"
-                  - !Join [",", !Ref WriteAccess]
+                - !Join
+                    - ","
+                    - - Fn::ImportValue: !Sub "${AWS::Region}-nextflow-forge-service-user-ServiceUserArn"
+                      - !Join
+                          - ","
+                          - !Ref WriteAccess
             Action:
               - "s3:*Object*"
               - "s3:*MultipartUpload*"
--- a/templates/nextflow-s3-template.yaml
+++ b/templates/nextflow-s3-template.yaml
@@ -18,35 +18,27 @@ Parameters:
     Type: String
     AllowedPattern: '^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$'
     ConstraintDescription: 'Must be an acceptable email address syntax(i.e. joe.smith@sagebase.org)'
-  EncryptBucket:
-    Type: String
-    # Description: true (default) to encrypt bucket, false for no encryption
-    # AllowedValues:
-    #  - true
-    #  - false
-    # Default: true
   BucketName:
     Type: String
     # Description: (Optional) Name of the created bucket.
     # Default: ""
-  GrantAccess:
+  BucketAccess:
     Type: CommaDelimitedList
     # Default: "[]"
     # Description: Grant bucket access to accounts, groups, and users.
     ConstraintDescription: List of ARNs (i.e. ["arn:aws:iam::011223344556:user/jsmith", "arn:aws:iam::544332211006:user/rjones"]
-  AllowWriteBucket:
-    Type: String
-    # Description: true for read-write, false (default) for read-only bucket
-    # AllowedValues:
-    #   - true
-    #   - false
-    # Default: false
+  ReadAccess:
+    Type: CommaDelimitedList
+    # Default: "[]"
+    # Description: Grant bucket access to accounts, groups, and users.
+    ConstraintDescription: List of ARNs (i.e. ["arn:aws:iam::011223344556:user/jsmith", "arn:aws:iam::544332211006:user/rjones"]
+  WriteAccess:
+    Type: CommaDelimitedList
+    # Default: "[]"
+    # Description: Grant bucket access to accounts, groups, and users.
+    ConstraintDescription: List of ARNs (i.e. ["arn:aws:iam::011223344556:user/jsmith", "arn:aws:iam::544332211006:user/rjones"]
 
 Conditions:
-  AllowWrite: !Equals [!Ref AllowWriteBucket, true]
-  AllowUserAccess: !Not [!Equals [!Join ['', !Ref GrantAccess], "[]"]]
-  EnableEncryption: !Equals [!Ref EncryptBucket, true]
-  DisableEncryption: !Not [!Condition EnableEncryption]
   HasBucketName: !Not [!Equals [!Ref BucketName, ""]]
 
 # Resources
@@ -62,7 +54,11 @@ Resources:
         # AllowedValues:
         #   - Enabled
         #   - Suspended
-    # Default: Suspended
+        # Default: Suspended
+      BucketEncryption:
+        ServerSideEncryptionConfiguration:
+          - ServerSideEncryptionByDefault:
+              SSEAlgorithm: AES256
 
   # Policy
   BucketPolicy:
@@ -73,19 +69,6 @@ Resources:
         Version: 2012-10-17
         Statement:
 
-          # - Sid: SynapseObjectAccess
-          #   # If Synapse is a grantee, gives Synapse access to objects in the bucket (R/O or R/W, depending on AllowWrite)
-          #   Effect: Allow
-          #   Principal:
-          #     AWS: !Ref GrantAccess
-          #   Action:
-          #     - !If [AllowWrite, "s3:*Object*", "s3:GetObject*"]
-          #     - "s3:*MultipartUpload*"
-          #   Condition:
-          #     StringEquals:
-          #       "aws:PrincipalAccount": "325565585839"
-          #   Resource: !Sub "arn:aws:s3:::${Bucket}/*"
-
           - Sid: DenyAllBucketAccess
             # denies access to everyone
             Effect: Deny
@@ -93,6 +76,10 @@ Resources:
               AWS: "*"
             Action:
               - "s3:*"
+            Condition:
+              StringEquals:
+                "aws:PrincipalAccount":
+                  - Fn::ImportValue: !Sub "${nextflow-forge-service-user}-ServiceUser"
             Resource:
               - !Sub arn:aws:s3:::${Bucket}/*
               - !Sub arn:aws:s3:::${Bucket}
@@ -101,10 +88,15 @@ Resources:
             # gives grantees access to the bucket
             Effect: Allow
             Principal:
-              AWS: !Ref GrantAccess
+              AWS: !Ref BucketAccess
             Action:
+              - "s3:ListBucket"
               - "s3:ListBucket*"
               - "s3:GetBucketLocation"
+            Condition:
+              StringEquals:
+                "aws:PrincipalAccount":
+                  - Fn::ImportValue: !Sub "${nextflow-forge-service-user}-ServiceUser"
             Resource:
               - !Sub arn:aws:s3:::${Bucket}/*
               - !Sub arn:aws:s3:::${Bucket}
@@ -113,12 +105,16 @@ Resources:
             # give grantees read access to objects
             Effect: Allow
             Principal:
-              AWS: !Ref GrantAccess
+              AWS: !Ref ReadAccess
             Action:
               - "s3:GetObject"
               - "s3:GetObjectAcl"
               - "s3:AbortMultipartUpload"
               - "s3:ListMultipartUploadParts"
+            Condition:
+              StringEquals:
+                "aws:PrincipalAccount":
+                  - Fn::ImportValue: !Sub "${nextflow-forge-service-user}-ServiceUser"
             Resource:
               - !Sub arn:aws:s3:::${Bucket}/*
               - !Sub arn:aws:s3:::${Bucket}
@@ -126,10 +122,14 @@ Resources:
           - Sid: WriteAccess
             Effect: Allow
             Principal:
-              AWS: !Ref GrantAccess
+              AWS: !Ref WriteAccess
             Action:
-              - !If [AllowWrite, "s3:*Object*", "s3:GetObject*"]
+              - "s3:*Object*"
               - "s3:*MultipartUpload*"
+            Condition:
+              StringEquals:
+                "aws:PrincipalAccount":
+                  - Fn::ImportValue: !Sub "${nextflow-forge-service-user}-ServiceUser"
             Resource:
               - !Sub arn:aws:s3:::${Bucket}/*
               - !Sub arn:aws:s3:::${Bucket}
--- a/templates/nextflow-s3-template.yaml
+++ b/templates/nextflow-s3-template.yaml
@@ -122,8 +122,7 @@ Resources:
               Principal:
                 AWS: !Ref ReadWriteAccessArns
               Action:
-                - "s3:AbortMultipartUpload"
-                - "s3:ListMultipartUploadParts"
+                - "s3:*MultipartUpload*"
                 - "s3:ListBucket*"
                 - "s3:GetBucketLocation"   
               Resource:
@@ -143,8 +142,6 @@ Resources:
                 - "s3:GetBucketLocation"
                 - "s3:GetObject"
                 - "s3:GetObjectAcl"
-                - "s3:AbortMultipartUpload"
-                - "s3:ListMultipartUploadParts"
               Resource:
                 - !Sub arn:aws:s3:::${Bucket}
                 - !Sub arn:aws:s3:::${Bucket}/*
--- a/templates/nextflow-s3-template.yaml
+++ b/templates/nextflow-s3-template.yaml
@@ -8,20 +8,20 @@ Parameters:
     Description: (Optional) Name of the created bucket.
   ReadWriteAccessArns:
     Type: CommaDelimitedList
-    Default: "[]"
+    Default: ""
     Description: (Optional) Grant read/write access to accounts, groups, and users.
     ConstraintDescription: List of ARNs (i.e. ["arn:aws:iam::011223344556:user/jsmith", "arn:aws:iam::544332211006:user/rjones"]
   ReadOnlyAccessArns:
     Type: CommaDelimitedList
-    Default: "[]"
+    Default: ""
     Description: (Optional) Grant read access to accounts, groups, and users.
     ConstraintDescription: List of ARNs (i.e. ["arn:aws:iam::011223344556:user/jsmith", "arn:aws:iam::544332211006:user/rjones"]
 
 # Conditions
 Conditions:
   HasBucketName: !Not [!Equals [!Ref BucketName, ""]]
-  HasReadWriteAccessArns: !Not [!Equals [!Join [",", !Ref ReadWriteAccessArns], "[]"]]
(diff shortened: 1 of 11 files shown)


tthyer (changes requested):
Good start. Please make the nextflow service user a default principal. In other words, don't pass it as a parameter, instead it should be a default principal that is joined to the other principal lists you're passing in. 

The stack that creates this user is nextflow-forge-service-user. You can view this in the AWS Cloudformation Console. See the outputs there -- you'll want the ExportName. Then you can use the Cloudformation `!ImportValue` function to grab that output value. [Here](https://github.com/Sage-Bionetworks-Workflows/aws-workflows-nextflow-infra/blob/main/templates/workflows-kms-key.yaml#L24) is an example of how to use ImportValue. Chat me if you have questions!

templates/nextflow-s3-template.yaml
```diff
@@ -0,0 +1,144 @@
...
+            Action:
+              - "s3:GetObject"
+              - "s3:GetObjectAcl"
+              - "s3:AbortMultipartUpload"
+              - "s3:ListMultipartUploadParts"
+            Resource:
+              - !Sub arn:aws:s3:::${Bucket}/*
+              - !Sub arn:aws:s3:::${Bucket}
+
+          - Sid: WriteAccess
```
zaro0508, line 121: does a principal need to be added to all three (BucketAccess, ReadAccess, and WriteAccess) to have write access to objects in this bucket? or does the principal only need to be in WriteAccess group to have write access to the bucket and objects?

templates/nextflow-s3-template.yaml
```diff
@@ -0,0 +1,144 @@
...
+              AWS: !Join
+                - ","
+                - - Fn::ImportValue: !Sub "${AWS::Region}-nextflow-forge-service-user-ServiceUserArn"
+                  - !Join [",", !Ref BucketAccess]
+            Action:
+              - "s3:ListBucket"
+              - "s3:ListBucket*"
+              - "s3:GetBucketLocation"
+            Resource:
+              - !Sub arn:aws:s3:::${Bucket}/*
```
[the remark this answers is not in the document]
tthyer: @daisyhan97 If it's not clear, what @zaro0508 means is that some of the S3 actions apply only at the bucket level and some apply only at the object level.

templates/nextflow-s3-template.yaml
daisyhan97: Just edited to allow users in WriteAccess to have permissions in BucketAccess and ReadAccess! So the principal only needs to be in the WriteAccess group to have access to the bucket and objects

templates/nextflow-s3-template.yaml
```diff
@@ -0,0 +1,139 @@
+AWSTemplateFormatVersion: 2010-09-09
+# Description: Custom Nextflow Bucket
```
tthyer, line 2: uncomment; I would name this something like "Private S3 bucket"

templates/nextflow-s3-template.yaml
```diff
@@ -0,0 +1,139 @@
...
+
+# Parameters
+Parameters:
+  Department:
+    # Description: 'The department for this resource'
```
tthyer, line 7: These also need to be uncommented. `Description` is a real parameter field.

templates/nextflow-s3-template.yaml
```diff
@@ -0,0 +1,139 @@
...
+    Type: String
+    AllowedPattern: '^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$'
+    ConstraintDescription: 'Must be an acceptable email address syntax(i.e. joe.smith@sagebase.org)'
+  BucketName:
+    Type: String
+    # Description: (Optional) Name of the created bucket.
+    # Default: ""
+  BucketAccess:
+    Type: CommaDelimitedList
+    # Default: "[]"
```
tthyer, line 27: I think it would be a good idea to use role-based access for this bucket. What I mean is that the current access is divided up by what a user can do -- can they list buckets? Can they read objects in a bucket? Can they write to a bucket? Instead I suggest that you think of it as three levels of role: 
1. Admin-level access has `s3:*` for this bucket and its objects -- this should be the CI user, root, and admin roles. I can help you set these. I think for this level you can set the principals directly rather than parameterize it.
2. Read-write access: they can list buckets, list objects, read objects, and write to the bucket. This would be an optional parameter, so it should have a Default value of empty (i.e. not required).
3. Read-only access: they can list and read. This would also be an optional parameter.

Let's discuss this more in person.

templates/nextflow-s3-template.yaml
```diff
@@ -0,0 +1,139 @@
...
+    Properties:
+      BucketName: !If [HasBucketName, !Ref BucketName, !Ref 'AWS::NoValue']
+      AccessControl: Private
+      # VersioningConfiguration:
+        # SType: String
+        # Description: Enabled to enable bucket versioning, default is Suspended
+        # AllowedValues:
+        #   - Enabled
+        #   - Suspended
+        # Default: Suspended
```
tthyer, line 57: I don't think we need versioning -- you can remove this.

templates/nextflow-s3-template.yaml
```diff
@@ -0,0 +1,139 @@
...
+            Effect: Allow
+            Principal:
+              AWS: !Split
+                - ","
+                - !Join
+                    - ","
+                    - - Fn::ImportValue: !Sub "${AWS::Region}-nextflow-forge-service-user-ServiceUserArn"
+                      - !Join
+                          - ","
+                          - !Ref WriteAccess
```
tthyer, line 120: looks good

tthyer (changes requested):

templates/nextflow-s3-template.yaml
```diff
@@ -0,0 +1,148 @@
...
+      AdminRole: "arn:aws:iam::035458030717:role/aws-reserved/sso.amazonaws.com/AWSReservedSSO_Administrator_580e9f32ac55c4e7"
+    "728882028485":
+      AdminRole: "arn:aws:iam::728882028485:role/aws-reserved/sso.amazonaws.com/AWSReservedSSO_Administrator_3a9d61d0884be260"
+
+# Resources
+Resources:
+  Bucket:
+    Type: AWS::S3::Bucket
+    Properties:
+      BucketName: !If [HasBucketName, !Ref BucketName, !Ref 'AWS::NoValue']
```
BrunoGrandePhD, line 52: Out of curiosity, can this be achieved by setting the default value of the `BucketName` parameter to `AWS::NoValue`? I'm not requiring that change here, but I'm curious so we can simplify templates going forward.

templates/nextflow-s3-template.yaml
```diff
@@ -0,0 +1,148 @@
...
+      AccessControl: Private
```
BrunoGrandePhD, line 53: I think we need to enable the `BucketOwnerPreferred` option ([example](https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/s3-bucket-v2.j2#L108-L110)) to ensure that the bucket policy applies to objects uploaded by other accounts. Similarly, we need to require that uploads to the bucket apply the `bucket-owner-full-control` canned ACL in the bucket policy still statement that enables uploads by external users/roles ([example](https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/s3-bucket-v2.j2#L233-L235)).

templates/nextflow-s3-template.yaml
```diff
@@ -0,0 +1,148 @@
...
+                - !Sub "arn:aws:iam::${AWS::AccountId}:root"
+                - Fn::ImportValue: !Sub "${AWS::Region}-workflows-nextflow-ci-service-account-ServiceRoleArn"
+                - !FindInMap [Accounts, !Ref AWS::AccountId, AdminRole]
+            Action:
+              - "s3:*"
+            Resource:
+              - !Sub arn:aws:s3:::${Bucket}
+              - !Sub arn:aws:s3:::${Bucket}/*
+
+          - Sid: ReadWriteAccess
```
BrunoGrandePhD, line 95: This would be the statement where you would need to add [this condition](https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/s3-bucket-v2.j2#L233-L235).

templates/nextflow-s3-template.yaml
```diff
@@ -0,0 +1,148 @@
...
+
+          - Sid: ReadOnlyAccess
+            # gives grantees read (only) access to the bucket
+            Effect: Allow
+            Principal:
+              AWS: !Split
+                - ","
+                - !Join
+                    - ","
+                    - - Fn::ImportValue: !Sub "${AWS::Region}-nextflow-forge-service-user-ServiceUserArn"
```
BrunoGrandePhD, line 124: Why are we including the Forge user on both the `ReadWriteAccess` and `ReadOnlyAccess` statements? Isn't it the latter redundant? Or is this being done just for simplicity in case no values are given to the `ReadOnlyAccess` parameter?

templates/nextflow-s3-template.yaml
```diff
@@ -0,0 +1,148 @@
...
+                      - !Join
+                          - ","
+                          - !Ref ReadOnlyAccess
```
BrunoGrandePhD, line 127: The bucket policy statement and template parameter share the same name, `ReadOnlyAccess`. I would avoid this by renaming both parameters as `ReadOnlyAccessArns` and `ReadWriteAccessArns`.

templates/nextflow-s3-template.yaml
```diff
@@ -0,0 +1,148 @@
...
+# Outputs
+Outputs:
+  BucketName:
+    Value: !Ref Bucket
+    Export:
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-BucketName'
+  BucketArn:
+    Value: !GetAtt Bucket.Arn
+    Export:
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-BucketArn'
```
BrunoGrandePhD, line 148: You're missing an empty line at the end of this file.

BrunoGrandePhD (changes requested):
Good work, @daisyhan97! I think the most important change that needs to be done is enabling `BucketOwnerPreferred` and enforcing the `BucketOwnerFullControl` canned ACL.

templates/nextflow-s3-template.yaml
daisyhan97: Forge user has been removed from ReadOnly! The !Join function doesn't work when the `ReadOnlyAccessArns` is left empty (in which case, should the Forge user be set as a default value for the ReadOnlyAccessArns parameter?)

templates/nextflow-s3-template.yaml
daisyhan97: The bucket-owner-full-control parameter only applies to the "s3:*Object*" parameter, so it was applied in `Sid: ReadWriteAccess`. Everything else required for ReadWrite access was added separately in `Sid: ReadWriteBucketAccess`.

templates/nextflow-s3-template.yaml
tthyer: No, because you'd have to use an intrinsic function for this, and they're not allowed in Parameters.

BrunoGrandePhD: Thanks for clarifying, @tthyer!

templates/nextflow-s3-template.yaml
```diff
@@ -23,18 +23,21 @@ Parameters:
...
+    Default: "[]"
+    Description: (Optional) Grant read access to accounts, groups, and users.
     ConstraintDescription: List of ARNs (i.e. ["arn:aws:iam::011223344556:user/jsmith", "arn:aws:iam::544332211006:user/rjones"]
 
 # Conditions
 Conditions:
   HasBucketName: !Not [!Equals [!Ref BucketName, ""]]
+  HasReadWriteAccessArns: !Not [!Equals [!Join [",", !Ref ReadWriteAccessArns], "[]"]]
+  HasReadOnlyAccessArns: !Not [!Equals [!Join [",", !Ref ReadOnlyAccessArns], "[]"]]
+  NoReadWriteAccessArns: !Equals [!Join [",", !Ref ReadWriteAccessArns], "[]"]
```
BrunoGrandePhD, line 40: Out of curiosity, can you refer to prior conditions in later conditions? If so, we could avoid some duplication using something like this: 

```suggestion
  NoReadWriteAccessArns: !Equals [!Join [",", !Ref ReadWriteAccessArns], "[]"]
  HasReadOnlyAccessArns: !Not [!Equals [!Join [",", !Ref ReadOnlyAccessArns], "[]"]]
  HasReadWriteAccessArns: !Not [!Ref NoReadWriteAccessArns]
```

BrunoGrandePhD: This change wouldn't be required if we dropped the `NoReadWriteAccessArns` condition as per my comment below.

templates/nextflow-s3-template.yaml
```diff
@@ -95,62 +98,89 @@ Resources:
...
-                          - !Ref ReadWriteAccessArns
-            Action:
-              - "s3:*Object*"  
-            Condition:
-              StringEquals:
-                s3:x-amz-acl: bucket-owner-full-control
-            Resource:
-              - !Sub arn:aws:s3:::${Bucket}/*    
+          - !If
+            - NoReadWriteAccessArns
```
BrunoGrandePhD, line 102: I think we can unconditionally grant read/write access to the Tower Forge IAM user since policy statements should complement each other. In other words, if two different statements grant two different principals the same permissions, they will both apply. So, we wouldn't need the `NoReadWriteAccessArns` condition.

templates/nextflow-s3-template.yaml
```diff
@@ -95,62 +98,89 @@ Resources:
...
+              # gives forge user read/write write access to the bucket if no read/write arns are supplied
+              Effect: Allow
+              Principal:
+                AWS:
+                  - Fn::ImportValue: !Sub "${AWS::Region}-nextflow-forge-service-user-ServiceUserArn"
+              Action:
+                - "s3:*Object*"  
+              Condition:
+                StringEquals:
+                  s3:x-amz-acl: bucket-owner-full-control
```
BrunoGrandePhD, line 113: We don't need this statement condition for the Tower Forge user because it's an IAM principal within the account and therefore the object ownership issue wouldn't apply if it uploaded objects to the bucket. In fact, this might break things because Nextflow probably doesn't use the `bucket-owner-full-control` ACL. So, it's actually a good thing that we have different statements for the Tower Forge user and other human users/roles.

templates/nextflow-s3-template.yaml
```diff
@@ -95,62 +98,89 @@ Resources:
...
+            - HasReadWriteAccessArns
+            - Sid: ReadWriteAccess
+              # gives grantees read/write write access to the bucket with required --acl
+              Effect: Allow
+              Principal:
+                AWS: !Split
+                  - ","
+                  - !Join
+                      - ","
+                      - - Fn::ImportValue: !Sub "${AWS::Region}-nextflow-forge-service-user-ServiceUserArn"
```
BrunoGrandePhD, line 128: Since the Tower Forge user doesn't need the `bucket-owner-full-control` condition, I would revert to granting `s3:*` to this role and only split off the `s3:*Object*` actions for the `ReadWriteAccessArns`.

templates/nextflow-s3-template.yaml
```diff
@@ -95,62 +98,89 @@ Resources:
...
+            - HasReadWriteAccessArns
+            - Sid: ReadWriteBucketAccess
+              # gives grantees with read/write access read access to the bucket
+              Effect: Allow
+              Principal:
+                AWS: !Split
+                  - ","
+                  - !Join
+                      - ","
+                      - - Fn::ImportValue: !Sub "${AWS::Region}-nextflow-forge-service-user-ServiceUserArn"
```
BrunoGrandePhD, line 151: Same as the above comment, you can simplify the list of principals by omitting the Forge user (since it'll have `s3:*` in another statement).

BrunoGrandePhD (changes requested):
I think we're almost there! I made a few suggestions that would greatly simplify the template. Good thinking on using those `If` statements as a workaround for the `Join` limitation.

templates/nextflow-s3-template.yaml
daisyhan97: Done! Added `Sid: ForgeUserFullAccess`

templates/nextflow-s3-template.yaml
```diff
@@ -0,0 +1,175 @@
...
+              AWS:
+                - Fn::ImportValue: !Sub "${AWS::Region}-nextflow-forge-service-user-ServiceUserArn"
+            Action:
+              - "s3:*"
+            Resource:
+              - !Sub arn:aws:s3:::${Bucket}
+              - !Sub arn:aws:s3:::${Bucket}/*
+
+          - !If
+            - HasReadWriteAccessArns
```
daisyhan97, line 113: @BrunoGrandePhD now that the !Join has been removed, should this condition be kept in?

templates/nextflow-s3-template.yaml
```diff
@@ -98,37 +97,26 @@ Resources:
...
-                - !Sub arn:aws:s3:::${Bucket}/*   
-            - !Ref AWS::NoValue 
+          - Sid: ForgeUserFullAccess
+            # gives forge user full access to bucket
+            Effect: Allow
+            Principal:
+              AWS:
+                - Fn::ImportValue: !Sub "${AWS::Region}-nextflow-forge-service-user-ServiceUserArn"
+            Action:
+              - "s3:*"
```
BrunoGrandePhD, line 107: On second thought, we could stick to what's in the [Tower docs](https://github.com/seqeralabs/nf-tower-aws/blob/master/launch/s3-bucket-write.json) (copied below). This way, we would be more conservative with permissions. You can probably combine the bucket-specific and object-specific statements into one as you did. 

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:ListBucket"
            ],
            "Resource": [
                "arn:aws:s3:::YOUR-BUCKET-NAME"
            ]
        },
        {
            "Action": [
                "s3:GetObject",
                "s3:PutObject",
                "s3:PutObjectTagging",
                "s3:DeleteObject"
            ],
            "Resource": [
                "arn:aws:s3:::YOUR-BUCKET-NAME/*"
            ],
            "Effect": "Allow"
        }
    ]
}
```

templates/nextflow-s3-template.yaml
daisyhan97: Nm - seems to break without

templates/nextflow-s3-template.yaml
```diff
@@ -0,0 +1,177 @@
...
+  Project:
+    Description: 'The name of the project that this resource is used for'
+    Type: String
+    AllowedPattern: '^\S*$'
+    ConstraintDescription: 'Must be string with no spaces'
+  OwnerEmail:
+    Description: 'Email address of the owner of this resource'
+    Type: String
+    AllowedPattern: '^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$'
+    ConstraintDescription: 'Must be an acceptable email address syntax(i.e. joe.smith@sagebase.org)'
```
BrunoGrandePhD, line 20: These parameters aren't being used. We can remove them and rely on `stack_tags` instead ([example](https://github.com/Sage-Bionetworks-Workflows/aws-workflows-nextflow-infra/blob/main/config/prod/nextflow-aurora-mysql.yaml#L10-L13)), which means you just need to delete these lines since `stack_tags` are specified in the config when the template is deployed. You can just accept my suggested edit below to do this quickly on GitHub. 

```suggestion
```

templates/nextflow-s3-template.yaml
```diff
@@ -0,0 +1,177 @@
...
+  BucketName:
+    Type: String
+    Description: (Optional) Name of the created bucket.
+  ReadWriteAccessArns:
+    Type: CommaDelimitedList
+    Default: "[]"
```
BrunoGrandePhD, line 26: You've confirmed that this should be quoted, right?

templates/nextflow-s3-template.yaml
```diff
@@ -0,0 +1,177 @@
...
+    Description: (Optional) Grant read/write access to accounts, groups, and users.
+    ConstraintDescription: List of ARNs (i.e. ["arn:aws:iam::011223344556:user/jsmith", "arn:aws:iam::544332211006:user/rjones"]
+  ReadOnlyAccessArns:
+    Type: CommaDelimitedList
+    Default: "[]"
```
BrunoGrandePhD, line 31: Same as above.

templates/nextflow-s3-template.yaml
```diff
@@ -0,0 +1,177 @@
...
+    Description: (Optional) Grant read access to accounts, groups, and users.
+    ConstraintDescription: List of ARNs (i.e. ["arn:aws:iam::011223344556:user/jsmith", "arn:aws:iam::544332211006:user/rjones"]
+
+# Conditions
+Conditions:
+  HasBucketName: !Not [!Equals [!Ref BucketName, ""]]
+  HasReadWriteAccessArns: !Not [!Equals [!Join [",", !Ref ReadWriteAccessArns], "[]"]]
+  HasReadOnlyAccessArns: !Not [!Equals [!Join [",", !Ref ReadOnlyAccessArns], "[]"]]
```
BrunoGrandePhD, line 39: I see you're using the quoted version of the square brackets here like your default values above. So, when you use `!Ref ReadWriteAccessArns` in the template below, this works as expected?

BrunoGrandePhD (approved):
Other than the suggested deletion, this looks good to go! 🚀

templates/nextflow-s3-template.yaml
daisyhan97: @BrunoGrandePhD Yes! If it's not quoted, it returns this error:
`"An error occurred (ValidationError) when calling the UpdateStack operation: Template format error: Every Default member must be a string."`

templates/nextflow-s3-template.yaml
daisyhan97: Yes! The Equals function returns an error if it isn't quoted:
`"An error occurred (ValidationError) when calling the UpdateStack operation: Template error: every Fn::Equals object requires a list of 2 string parameters."`

templates/nextflow-s3-template.yaml
BrunoGrandePhD: It seems like `[]` is just a flag value for indicating that nothing was provided and is not being interpreted as an empty list. 

I would try dropping the two `Default: "[]"` lines from this template and replace `"[]"` in the conditions with the empty string `""`. If that doesn't work, then I would add back `Default: ""`.

daisyhan97: Dropping the Default makes the parameter required. Added `Default: ""`

templates/nextflow-s3-template.yaml
```diff
@@ -0,0 +1,162 @@
...
+            Effect: Allow
+            Principal:
+              AWS:
+                - Fn::ImportValue: !Sub "${AWS::Region}-nextflow-forge-service-user-ServiceUserArn"
+            Action:
+              - "s3:ListBucket"
+              - "s3:GetObject"
+              - "s3:PutObject"
+              - "s3:PutObjectTagging"
+              - "s3:DeleteObject"
```
tthyer, line 96: I like this -- using principle of minimum permissions for this user.

templates/nextflow-s3-template.yaml
```diff
@@ -0,0 +1,162 @@
...
+              # gives grantees read (only) access to the bucket
+              Effect: Allow
+              Principal:
+                AWS: !Ref ReadOnlyAccessArns
+              Action:
+                - "s3:ListBucket*"
+                - "s3:GetBucketLocation"
+                - "s3:GetObject"
+                - "s3:GetObjectAcl"
+                - "s3:AbortMultipartUpload"
```
tthyer, line 146: Why give AbortMultipartUpload to read-only user?

tthyer (approved):
I had one question but overall I think this is ready.

tthyer (approved):

templates/nextflow-s3-template.yaml
daisyhan97: ListMultipartUpload and AbortMultipartUpload were given to the read-only user [here](https://github.com/Sage-Bionetworks/aws-infra/blob/8e98fb5727027df0b22bf688f21b49c254391758/templates/s3-bucket-v2.j2#L196) -- but can be removed if they aren't needed in this case!

tthyer: Perhaps @zaro0508 can shed light on why that permission is given to read-only

zaro0508: @daisyhan97 The policy you referenced was crafted by @brucehoff specifically for users who want direct access  to [synapse custom storage locations](https://help.synapse.org/docs/Custom-Storage-Locations.2048327803.html).  I'm not sure if you are trying to accomplish the same use case here but all info and test for that use case are in issue IT-1295

daisyhan97: Thank you for clarifying, @zaro0508!
@tthyer I wasn't able to find the reasoning behind the MultipartUpload permissions in IT-1295, so I removed both the MultipartUpload permissions from the ReadOnly users!

tthyer:
thanks @daisyhan97, this looks good.

PR #37, opened 2021-08-11
WORKFLOWS-36 use custom keys

tthyer:
[WORKFLOWS-36](https://sagebionetworks.jira.com/browse/WORKFLOWS-36)

- add admins to workflows-infra key
- use in aurora stack and new tower secrets stack for encryption

### DIFF
--- a/config/develop/nextflow-aurora-mysql.yaml
+++ b/config/develop/nextflow-aurora-mysql.yaml
@@ -1,12 +1,14 @@
 template_path: nextflow-aurora-mysql.yaml
 stack_name: nextflow-aurora-mysql
 dependencies:
+  - common/workflows-kms-key.yaml
   - develop/nextflow-vpc.yaml
 parameters:
   VpcID: !stack_output_external nextflow-vpc::VPCId
   SubnetIDs:
     - !stack_output_external nextflow-vpc::PrivateSubnet1
     - !stack_output_external nextflow-vpc::PrivateSubnet2
+  KmsKeyAlias: !stack_output_external workflows-infra-kms-key::Alias
 stack_tags:
   Department: IBC
   Project: Infrastructure
--- a/templates/workflows-kms-key.yaml
+++ b/templates/workflows-kms-key.yaml
@@ -8,7 +8,15 @@ Parameters:
     Description: URL of S3 bucket where templates are deployed
     ConstraintDescription: Must be a valid S3 HTTP URL
 
+Mappings:
+  Accounts:
+    "035458030717":
+      AdminRole: "arn:aws:iam::035458030717:role/aws-reserved/sso.amazonaws.com/AWSReservedSSO_Administrator_580e9f32ac55c4e7"
+    "728882028485":
+      AdminRole: "arn:aws:iam::728882028485:role/aws-reserved/sso.amazonaws.com/AWSReservedSSO_Administrator_3a9d61d0884be260"
+
 Resources:
+
   WorkflowsKeyStack:
     Type: AWS::CloudFormation::Stack
     Properties:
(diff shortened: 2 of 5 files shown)


templates/workflows-kms-key.yaml
```diff
@@ -8,7 +8,15 @@ Parameters:
     Description: URL of S3 bucket where templates are deployed
     ConstraintDescription: Must be a valid S3 HTTP URL
 
+Mappings:
+  Accounts:
+    "035458030717":
+      AdminRole: "arn:aws:iam::035458030717:role/aws-reserved/sso.amazonaws.com/AWSReservedSSO_Administrator_580e9f32ac55c4e7"
+    "728882028485":
+      AdminRole: "arn:aws:iam::728882028485:role/aws-reserved/sso.amazonaws.com/AWSReservedSSO_Administrator_3a9d61d0884be260"
```
zaro0508, line 16: There is only one set of SSO user accounts and roles.  All SSO user account and roles used to access nextflow accounts (dev & prod) are in the our AWS organizations account (531805629419) therefore this should be.. 

```
    "035458030717":
      AdminRole: "arn:aws:iam::531805629419:role/aws-reserved/sso.amazonaws.com/AWSReservedSSO_Administrator_580e9f32ac55c4e7"
    "728882028485":
      AdminRole: "arn:aws:iam::531805629419:role/aws-reserved/sso.amazonaws.com/AWSReservedSSO_Administrator_3a9d61d0884be260"
```

Also Are you sure you want admins in the nextflow dev account to have access to keys that will unlock data in your nextflow prod account?  That's probably not a good idea, just want to to make sure because I think this is what's going to happen here.

templates/workflows-kms-key.yaml
```diff
@@ -21,8 +29,12 @@ Resources:
           - - !Sub 'arn:aws:iam::${AWS::AccountId}:root'
             - !ImportValue
               'Fn::Sub': '${AWS::Region}-workflows-nextflow-ci-service-account-ServiceUserArn'
```
zaro0508, line 31: service user should probably only be allowed to `use` the key not admin it?

config/develop/nextflow-aurora-mysql.yaml
```diff
@@ -1,12 +1,14 @@
 template_path: nextflow-aurora-mysql.yaml
 stack_name: nextflow-aurora-mysql
 dependencies:
+  - common/workflows-kms-key.yaml
```
zaro0508, line 4: doesn't `nextflow-tower-secret.yaml` need to exist before running this template?

zaro0508 (changes requested):

templates/workflows-kms-key.yaml
tthyer: @zaro0508 Doesn't the server user have to be able to admin it in order to create and delete through CI?

config/develop/nextflow-aurora-mysql.yaml
tthyer: No, those secrets are not use in the aurora stack. It has its own.

templates/workflows-kms-key.yaml
tthyer: Oh, thanks for the correct of the account to use in the role ARNs, as I hadn't noticed.

Re "Are you sure you want admins in the nextflow dev account to have access to keys that will unlock data in your nextflow prod account" <-- that makes no sense since we're using `FindInMap` based on account id.

zaro0508: please ignore my comment, `There is only one set of SSO user accounts and roles`. it is not correct. there is one set of users but i forgot that SSO duplicates roles to each account that users access therefore your map is correct

templates/workflows-kms-key.yaml
zaro0508: ahh, yes you are correct. Service user needs access to update it once it creates it.

zaro0508 (approved):

PR #39, opened 2021-08-12
Workflows-22 ECS task definition

tthyer:
This creates the ECS task definition which will actually deploy the Nextflow application into the ECS cluster waiting for it.
This template is nearly the same as the one supplied by Seqera. Minor edits were required to make this work. I could make a PR against their private repo later to take these changes into account, however, further changes will be necessary in a follow-up PR to implement the load balancer as well.

Depends on #37


BrunoGrandePhD (approved):
🚀

PR #41, opened 2021-08-17
KMS key for bootstrapping must be in same stack with Lambda

tthyer:
Fix problem with kms key use in Aurora. The build is currently broken b/c the lambda role is attempting to use the workflows-infra key, but it is not a principal of that key.

### DIFF
--- a/templates/nextflow-aurora-mysql.yaml
+++ b/templates/nextflow-aurora-mysql.yaml
@@ -47,10 +47,10 @@ Parameters:
     Type: String
     Default: ''
 
-  KmsKeyAlias:
-    Description: Optional. KMS key alias used to encrypt secrets stored in Secrets Manager.
+  TemplateRootUrl:
     Type: String
-    Default: 'alias/workflows-infra'
+    Description: URL of S3 bucket where templates are deployed
+    ConstraintDescription: Must be a valid S3 HTTP URL
 
 Mappings:
   ClusterSettings:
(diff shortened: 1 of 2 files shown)


templates/nextflow-aurora-mysql.yaml
```diff
@@ -47,24 +47,50 @@ Parameters:
...
   ClusterSettings:
     mysql:
       version: 5.7.mysql_aurora.2.07.1
       engine: aurora-mysql
       family: aurora-mysql5.7
+  Accounts:
+    "035458030717":
+      AdminRole: "arn:aws:iam::035458030717:role/aws-reserved/sso.amazonaws.com/AWSReservedSSO_Administrator_580e9f32ac55c4e7"
+    "728882028485":
+      AdminRole: "arn:aws:iam::728882028485:role/aws-reserved/sso.amazonaws.com/AWSReservedSSO_Administrator_3a9d61d0884be260"
```
tthyer, line 65: Yet another place I am adding this mapping.
Really wish there were a generic way to refer to the Admin role. 

There are a couple ways this could be solved:
1. For sceptre configs that are in develop and prod, just add an AdminRole parameter to be passed in. This could use a stack_config value.
2. Add this snippet into an S3 bucket and include it in templates as an AWS::Include. This may be a good option if we continue putting configs in common. 
3. It would be great if sceptre had a way to dynamically determine the admin roles so that this configuration could be moved out of the templates and into the config. I.e., in CFN templates one can use the pseudo-parameter `AWS::AccountId` but afaik sceptre is unaware of that value and it's not something I want to pass as a sceptre `--var`.

BrunoGrandePhD: The first option sounds good to me. If I understand correctly, we would add `config.yaml` files in `prod/` and `dev/`, which would include the account-specific admin roles. This implies that any template that uses this value can't be in `common/`. I'm fine with this since it still seems more maintainable than adding a mapping in each template.

BrunoGrandePhD (approved):
Looks good to me! 

I noticed that the template root URL uses S3 instead of GitHub. Should we use S3 everywhere then?

templates/nextflow-aurora-mysql.yaml
zaro0508: i believe you can also setup a variable file and add pass this dictionary to every sceptre execution.  https://sceptre.cloudreach.com/2.6.3/docs/cli.html#variable-handling

PR #42, opened 2021-08-18
Add ACM certificates for subdomains

tthyer:
Acc. to various slack conversations as documented in [WORKFLOWS-40](https://sagebionetworks.jira.com/browse/WORKFLOWS-40), we have to create separate certificates for our new subdomains.

### DIFF
--- a/config/develop/tower-dev-certificate.yaml
+++ b/config/develop/tower-dev-certificate.yaml
@@ -1,0 +1,11 @@
+template_path: remote/acm-certificate.yaml
+stack_name: tower-dev-certificate
+parameters:
+  Department: IBC
+  Project: Infrastructure
+  OwnerEmail: nextflow-admins@sagebase.org
+  DnsDomainName: sagebionetworks.org
+  DnsSubDomainName: tower-dev
+hooks:
+  before_launch:
+    - !cmd "wget {{stack_group_config.aws_infra_templates_root_url}}/v0.3.0/templates/acm-certificate.yaml -O templates/remote/acm-certificate.yaml"
--- a/config/prod/tower-certificate.yaml
+++ b/config/prod/tower-certificate.yaml
@@ -1,0 +1,11 @@
+template_path: remote/acm-certificate.yaml
+stack_name: tower-certificate
+parameters:
+  Department: IBC
+  Project: Infrastructure
+  OwnerEmail: nextflow-admins@sagebase.org
+  DnsDomainName: sagebionetworks.org
+  DnsSubDomainName: tower
+hooks:
+  before_launch:
+    - !cmd "wget {{stack_group_config.aws_infra_templates_root_url}}/v0.3.0/templates/acm-certificate.yaml -O templates/remote/acm-certificate.yaml"


config/develop/tower-dev-certificate.yaml
```diff
@@ -0,0 +1,11 @@
...
+stack_name: tower-dev-certificate
+parameters:
+  Department: IBC
+  Project: Infrastructure
+  OwnerEmail: nextflow-admins@sagebase.org
+  DnsDomainName: sagebionetworks.org
+  DnsSubDomainName: tower-dev
+hooks:
+  before_launch:
+    - !cmd "wget {{stack_group_config.aws_infra_templates_root_url}}/v0.3.0/templates/acm-certificate.yaml -O templates/remote/acm-certificate.yaml"
```
zaro0508, line 11: hmm, i'm not a fan of this template anymore since it's setting Dept, project, and OwnerEmail tags.  would prefer to see `stack_tags: ..` here instead.  Maybe we should create a new file in `Sage-Bionetworks/aws-infra/ACM/certificate.yaml` ?

config/develop/tower-dev-certificate.yaml
```diff
@@ -0,0 +1,11 @@
+template_path: remote/acm-certificate.yaml
+stack_name: tower-dev-certificate
+parameters:
+  Department: IBC
+  Project: Infrastructure
+  OwnerEmail: nextflow-admins@sagebase.org
+  DnsDomainName: sagebionetworks.org
```
zaro0508, line 7: hmm, isn't this domain already registered?  I think you need to setup your own unique domain then redirect to it from sagebionetworks.org

zaro0508 (changes requested):
You'll need @xschildw help with this because creating a cert requires manual validation at sagebionetworks.org.  Unfortunately i don't know where that's registered at.

config/develop/tower-dev-certificate.yaml
tthyer: Sorry -- I am overwhelmed with other work. I will happily use an updated template in aws-infra but this is the available one and that it's not using stack_tags does not inhibit its functionality.

config/develop/tower-dev-certificate.yaml
tthyer: @zaro0508 Do you see the `DnsSubDomainName` directly underneath it? These two are being joined together. This is what the template contents looks like that does that:
```
  Certificate:
    Type: AWS::CertificateManager::Certificate
    Properties:
      DomainName: !Join
        - '.'
        - - !Ref DnsSubDomainName
          - !Ref DnsDomainName
      DomainValidationOptions:
        - DomainName: !Join
            - '.'
            - - !Ref DnsSubDomainName
              - !Ref DnsDomainName
          ValidationDomain: !Ref DnsDomainName
      ValidationMethod: "DNS"
```

tthyer: @xschildw Can you verify that I am setting this up correctly?

xschildw: Looks good, wondering about that ValidationDomain option in the template itself...

tthyer: The [doc](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-certificatemanager-certificate-domainvalidationoption.html) says that the `ValidationDomain` is:

> The domain name to which you want ACM to send validation emails. This domain name is the suffix of the email addresses that you want ACM to use. This must be the same as the DomainName value or a superdomain of the DomainName value. For example, if you request a certificate for testing.example.com, you can specify example.com as this value. In that case, ACM sends domain validation emails to the following five addresses:
> 
> admin@example.com
> 
> administrator@example.com
> 
> hostmaster@example.com
> 
> postmaster@example.com
> 
> webmaster@example.com

Sounds pretty benign

thomasyu888 (approved):

tthyer:
@zaro0508 I think the changes you are requesting are that I improve the aws-infra template. While I agree that would be a nice to have, I don't agree that improving a template provided by IT that does actually work fine should block progress on this project. If there were other changes you required, let me know and I will re-open this PR.

PR #46, opened 2021-08-20
[WORKFLOWS-15/34] Create README with onboarding docs

BrunoGrandePhD:
Here's a first draft. Does anyone have feedback? Are there any other sections that I should add? 

The formatted README can be read [**here**](https://github.com/Sage-Bionetworks-Workflows/aws-workflows-nextflow-infra/blob/bgrande/workflows-15/update-readme/README.md).

PR #47, opened 2021-08-20
Update LICENSE

thomasyu888:


BrunoGrandePhD (commented):
Based on the previous paragraph, I don't think this line is supposed to be changed. This is just stating the format. The [Spark license](https://github.com/apache/spark/blob/master/LICENSE#L189) also left this line as is. I think we need to include `Copyright 2021 Sage Bionetworks` in other files, but I just included it in the README.

thomasyu888:
Hmm, I read that last paragraph differently... 

> To apply the Apache License to your work, attach the following
      boilerplate notice, with the fields enclosed by brackets "[]"
      replaced with your own identifying information. (Don't include
      the brackets!)

Does that mean to replace the values in the [] with our own identifying info?

BrunoGrandePhD:
@thomasyu888: I read that same paragraph a few times when working on #46 because it was confusing. In particular, this phrase:

> To apply the Apache License to your work, attach the following boilerplate notice

Ultimately, my interpretation was that we needed to attach `Copyright 2021 Sage Bionetworks` to our work, which doesn't include the license itself.

PR #48, opened 2021-08-24
Nextflow ECS Service and ALB

tthyer:
Primarily:
- Adds ECS Service
- Adds application load balancer

Secondarily:
- Removes nextflow-ecs-capacity stack in favor of ApplicationAutoscaling resources in service stack
- Move security group out of ecs cluster stack to decouple aurora stack from ecs cluster stack
- Adds a Route 53 record set
- Parameterizes values for easier updates and sharing values between stacks


config/develop/nextflow-ecs-cluster.yaml
```diff
@@ -1,13 +1,18 @@
...
     - !stack_output_external nextflow-vpc::PublicSubnet
-  SecurityIngressFromPort: '80'
-  SecurityIngressToPort: '8080'
+sceptre_user_data:
+  TowerConfigFileContents: |
+    tower:
+      auth:
+        github:
+          allow-list:
+            - tess.thyer@gmail.com
```
tthyer, line 15: This list will be built out more later -- this configuration does not appear to be taking hold. Waiting to hear back from support about this.

templates/nextflow-aurora-mysql.yaml
```diff
@@ -166,6 +170,12 @@ Resources:
     Properties:
       VpcId: !Ref VpcID
       GroupDescription: Aurora Cluster Security Group
+      SecurityGroupIngress:
+        - Description: Inbound rule to allow database access to ECS cluster
+          SourceSecurityGroupId: !Ref EcsSecurityGroupId
+          IpProtocol: tcp
+          FromPort: 3306
+          ToPort: 3306
```
tthyer, line 178: Thanks @BrunoGrandePhD for pointing out I'd missed this

templates/nextflow-ecs-cluster.j2
```diff
@@ -135,17 +107,35 @@ Resources:
...
+          write_files:
+          - path: /opt/nextflow/tower.yml
+            content: |
+              {{ sceptre_user_data.TowerConfigFileContents|indent(14) }}
+            permissions: '0755'
+          - path: /etc/ecs/ecs.config
+            content: |
+              ECS_CLUSTER=${EcsClusterName}
+              ECS_BACKEND_HOST=
+              ECS_IMAGE_PULL_BEHAVIOR=prefer-cached
```
tthyer, line 137: Note this is the setting I discovered that we need in ECS last year. Got rate-limited again while testing this infrastructure -- it was crashing, so kept trying to re-pull all the images.

templates/nextflow-ecs-service.yaml
```diff
@@ -0,0 +1,209 @@
...
+      - PolicyName: service-autoscaling
+        PolicyDocument:
+          Statement:
+          - Effect: Allow
+            Action:
+              - application-autoscaling:*
+              - cloudwatch:DescribeAlarms
+              - cloudwatch:PutMetricAlarm
+              - ecs:DescribeServices
+              - ecs:UpdateService
```
BrunoGrandePhD, line 172: Out of curiosity, how do you get these specific lists of actions? Are they usually documented somewhere? Or do you pull these from examples?

BrunoGrandePhD (approved):
I did my best reviewing this PR since there are a lot of resources that I haven't used or written CFN templates for. It looks good to me! 🚀 

I only had a few questions for my own personal curiosity.

templates/nextflow-ecs-service.yaml
```diff
@@ -0,0 +1,209 @@
...
+    Type: AWS::EC2::SecurityGroup
+    Properties:
+      GroupName: !Join
+        - '-'
+        - - !Ref 'AWS::StackName'
+          - EcsAlbSecurityGroup
+      GroupDescription: Security Group for Nextflow ECS load balancer
+      VpcId: !Ref VpcId
+
+  EcsAlbSecurityGroupIngress:
```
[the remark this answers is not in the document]
tthyer: I find this to be cleaner, that's all

templates/nextflow-ecs-service.yaml
```diff
@@ -0,0 +1,209 @@
...
+    Type: AWS::EC2::SecurityGroupIngress
+    Properties:
+      GroupId: !GetAtt EcsAlbSecurityGroup.GroupId
```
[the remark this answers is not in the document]
tthyer: There is an important difference b/tw `Ref` and `GetAtt`. Most CFN resources have some default value they return when you use the `Ref` function. For example, the security group documentation indicates that `Ref` returns the resource id rather than the group id. Many CFN resources allow one to access additional fields -- this is always through `GetAtt`. For a security group the additional fields are the GroupId and the VpcId.

See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group.html#aws-properties-ec2-security-group-return-values-ref

templates/nextflow-ecs-service.yaml
tthyer: In this case I was working from an [AWS template snippet ](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/quickref-ecs.html)as a starting point.

templates/nextflow-ecs-task-definition.yaml
```diff
@@ -46,29 +43,89 @@ Parameters:
...
+    Description: The client secret provided by GitHub for the OAuth app
+    NoEcho: 'true'
+  RedisContainerName:
+    Type: String
+    Description: (Optional) Name of the container that runs redis
+    Default: redis
+  RedisContainerImage:
+    Type: String
+    Description: (Optional) Redis container docker image
+    Default: 'redis:5.0.8'
```
zaro0508, line 61: seems better to make these container image strings required params and explicitly pass in the image names because you will probably want to eventually update and make them different for each environment.

templates/nextflow-ecs-task-definition.yaml
```diff
@@ -173,16 +234,33 @@ Resources:
...
-  TowerService:
-    Type: AWS::ECS::Service
-    Properties:
-      Cluster: !Ref 'ClusterName'
-      DesiredCount: 1
-      ServiceName: TowerService
-      TaskDefinition: !Ref 'TowerTask'
-      LaunchType: EC2
+          MountPoints:
+            - ContainerPath: '/tower.yml'
```
zaro0508, line 247: `tower.yml` appears in other files and in this file a few times.  would be nice to make it a parameter so you can make it consistent.

templates/nextflow-ecs-service.yaml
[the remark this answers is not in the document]
BrunoGrandePhD: Oh, I was under the impression that in this case, the resource ID returned by `Ref` would be the security group ID.

templates/nextflow-ecs-task-definition.yaml
```diff
@@ -46,29 +43,89 @@ Parameters:
     Type: String
     Description: MySQL DB password
     NoEcho: 'true'
+  TowerGithubClientId:
+    Type: String
+    Description: The client id provided by GitHub for the OAuth app
+    NoEcho: 'true'
```
zaro0508, line 49: you'll probably want a `NoEcho: 'true'` for TowerDbUser as well.

zaro0508 (changes requested):
phew, that was a wopper!

config/develop/nextflow-ecs-cluster.yaml
```diff
@@ -1,13 +1,18 @@
...
   - develop/nextflow-vpc.yaml
 parameters:
-  VpcId: !stack_output_external nextflow-vpc::VPCId
+  EcsSecurityGroupId: !stack_output_external nextflow-ecs-security-group::SecurityGroupId
   SubnetIds:
     - !stack_output_external nextflow-vpc::PublicSubnet
-  SecurityIngressFromPort: '80'
-  SecurityIngressToPort: '8080'
+sceptre_user_data:
+  TowerConfigFileContents: |
```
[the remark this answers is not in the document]
tthyer: This is being written out as a string through cloud-init write_files module and it works fine. Using `!file` did not work and caused the dict error that I described to you in Slack.

Regardless, in a follow-up PR this config file will be moved out of these templates and into independent ones, so I dont' want to change what's working now.

templates/nextflow-ecs-cluster.j2
```diff
@@ -135,17 +107,35 @@ Resources:
       AssociatePublicIpAddress: true
       IamInstanceProfile: !Ref EcsInstanceProfile
       SecurityGroups:
-        - !Ref EcsSecurityGroup
+        - !Ref EcsSecurityGroupId
```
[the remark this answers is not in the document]
tthyer: yes

templates/nextflow-ecs-cluster.j2
```diff
@@ -135,17 +107,35 @@ Resources:
...
+
+          --==BOUNDARY==
+          Content-Type: text/cloud-config; charset="us-ascii"
+
+          #cloud-config
+          repo_update: true
+          repo_upgrade: security
+
+          write_files:
+          - path: /opt/nextflow/tower.yml
```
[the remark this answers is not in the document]
tthyer: because that's what the nextflow tower app requires

templates/nextflow-ecs-security-group.yaml
```diff
@@ -0,0 +1,44 @@
+AWSTemplateFormatVersion: 2010-09-09
+
+Description: AWS resources shared by ECS cluster and Aurora DB cluster
```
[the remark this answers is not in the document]
tthyer: yes, was originally intended to hold more but I think it will be limited to the one resource type

templates/nextflow-ecs-cluster.j2
```diff
@@ -135,17 +107,35 @@ Resources:
...
+            content: |
+              {{ sceptre_user_data.TowerConfigFileContents|indent(14) }}
+            permissions: '0755'
+          - path: /etc/ecs/ecs.config
+            content: |
+              ECS_CLUSTER=${EcsClusterName}
+              ECS_BACKEND_HOST=
```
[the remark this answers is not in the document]
tthyer: This is commonly seen in ecs.config files; it's a placeholder if you want to set the ecs endpoint. It's not necessary and I will remove to avoid confusion.

templates/nextflow-ecs-task-definition.yaml
tthyer: Good suggestion -- I will do this in the follow up PR for a branch already in progress that adds EFS

zaro0508 (approved):

PR #50, opened 2021-08-26
[WORKFLOWS-43] Create comprehensive stack for compute environments

BrunoGrandePhD:
Here's a summary of the changes introduced in this PR: 

- The Tower Forge service user credentials are now stored into Secrets Manager instead of being a CloudFormation output
- The Tower Forge service user credentials are encrypted using a narrowly scoped KMS key
- IAM roles are created for accessing the service user credentials in Secrets Manager and the S3 bucket
- Separate roles are created for read-only access and read/write-access to the S3 bucket
- An `AdminArns` parameter was added to move account-specific ARNs from the template to the Sceptre configs
- The S3 bucket policy is configured to deny access to all principals other than those listed (_i.e._ the access and admin roles)


config/prod/demo-project.yaml
```diff
@@ -0,0 +1,22 @@
+template_path: tower-project.yaml
+stack_name: demo-project
+
+parameters:
+  ReadWriteAccessArns:
+    - arn:aws:sts::035458030717:assumed-role/AWSReservedSSO_Viewer_fd80909e6a51c6e7/bruno.grande@sagebase.org
+  ReadOnlyAccessArns:
+    - arn:aws:sts::035458030717:assumed-role/AWSReservedSSO_Viewer_fd80909e6a51c6e7/tess.thyer@sagebase.org
+  AdminRoleArns:
```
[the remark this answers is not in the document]
BrunoGrandePhD: I was able to get this to work. Thanks for the tip!

tthyer (approved):

templates/tower-project.yaml
```diff
@@ -0,0 +1,368 @@
...
+                  - !Join
+                    - "|"
+                    - - !Join
+                        - "|"
+                        - !Ref ReadWriteAccessArns
+                      - !Join
+                        - "|"
+                        - !Ref AdminIds
+            Action:
+              - "sts:AssumeRole"
```
zaro0508, line 128: hmm, i'm not sure i like this idea.  This allows JC viewers to assume a role that provides full S3 access.  This seems like the same security problem that we discussed in slack.  For example when SageIT assigns a JC user to a nextflow viewer group we expect that they only have viewer access.  If you do this then that expectation becomes irrelevant.  Basically our JC configurations will have no meaning.  The source of truth for access should be JC user groups.

templates/tower-project.yaml
```diff
@@ -0,0 +1,368 @@
...
+                  - !Join
+                    - "|"
+                    - - !Join
+                        - "|"
+                        - !Ref ReadOnlyAccessArns
+                      - !Join
+                        - "|"
+                        - !Ref AdminIds
+            Action:
+              - "sts:AssumeRole"
```
zaro0508, line 153: same here.

zaro0508 (changes requested):

templates/tower-compute-environment.j2
```diff
@@ -0,0 +1,283 @@
...
+        Version: "2012-10-17"
+        Statement:
+          - Effect: "Allow"
+            Principal:
+              AWS: !Ref ReadOnlyAccessArns
+            Action:
+              - "sts:AssumeRole"
+
+  EncryptionKeyStack:
+    Type: AWS::CloudFormation::Stack
```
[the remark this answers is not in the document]
zaro0508: This is my suggestion because i feel that templates are more maintainable when not using nested stacks however that is only a preference, one that we discuss further in slack.  You can take my suggestion or do as you wish.

templates/tower-project.yaml
BrunoGrandePhD: We don't want the JC role on the production account to have anything more than `ViewOnlyAccess`, but we also need the ability to grant specific individuals with write-access to the resources they deployed using this template. 

Admittedly, granting write-access to a role called `Viewer` is counter-intuitive. To avoid any surprises, I propose that we can get rid of the `Viewer` role on our production account and deploy the `TowerUser` role described in Sage-Bionetworks-IT/organizations-infra#278. I updated the [role description](https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/278/files#diff-c89df69681d7714e12cc8907e3d7e5b10a9bee477bbff86cd269b81ad0c7a1b5R519-R521) so that the intended use is clear:

> SSO: View-only role that is also used for assuming project-specific roles with additional access

This way, it's clear to Sage IT that adding a user to the `aws-workflow-nextflow-tower-user` JC group not only grant view-access, but also the opportunity to assume a separate role with additional resource-specific write-access. 

If you think this is a fair approach, you can go ahead and approve this PR and Sage-Bionetworks-IT/organizations-infra#278.

tthyer: I like this approach

templates/tower-project.yaml
```diff
@@ -0,0 +1,338 @@
...
+    Type: List<String>
+    Description: >-
+      List of IAM user and role ARNs (strings) who should have access to 
+      project-specific resources (e.g., private S3 bucket, KMS key, secrets). 
+  
+  NonAdminArns:
+    Type: List<String>
+    Description: >-
+      List of IAM user and role ARNs (strings) who shouldn't have access to 
+      project-specific resources (e.g., private S3 bucket, KMS key, secrets).
```
BrunoGrandePhD, line 65: I confirmed that I don't have to explicitly deny the non-admin ARNs on the resource policy for the secret.

templates/tower-project.yaml
```diff
@@ -0,0 +1,338 @@
...
+              - HasReadOnlyAccessArns
+              - !Join
+                - ","
+                - !Ref ReadOnlyAccessArns
+              - !Ref AWS::NoValue
+
+  TowerBucket:
+    Type: AWS::S3::Bucket
+    Properties:
+      BucketName: !Sub "${AWS::StackName}-tower-bucket"
```
BrunoGrandePhD, line 124: I'm setting the bucket name because users can no longer use `aws s3 ls` to find the name (since they no longer have `s3:ListAllMyBuckets`).

templates/tower-project.yaml
```diff
@@ -0,0 +1,338 @@
...
+              Resource:
+                - !Sub "arn:aws:s3:::${TowerBucket}"
+                - !Sub "arn:aws:s3:::${TowerBucket}/*"
+
+  TowerForgeServiceUserAccessKey:
+    Type: AWS::IAM::AccessKey
+    Properties:
+      UserName: !Ref TowerForgeServiceUser
+
+  TowerProjectConfigurationSecret:
```
BrunoGrandePhD, line 231: I'm now storing various configuration details in this secret since users will no longer have access to CloudFormation output values (due to the fact that I can't bind a resource-based policy to a CFN stack).

templates/tower-project.yaml
```diff
@@ -0,0 +1,338 @@
...
+    Type: AWS::SecretsManager::Secret
+    Properties:
+      Name: !Sub "${AWS::StackName}-TowerProjectConfiguration"
```
BrunoGrandePhD, line 234: I have to set a name here so that that secret ARN is predictable for users.

templates/tower-project.yaml
```diff
@@ -0,0 +1,338 @@
...
+                      - ","
+                      - !Ref ReadWriteAccessArns
+                    - !Ref AWS::NoValue
+                  - !If
+                    - HasReadOnlyAccessArns
+                    - !Join
+                      - ","
+                      - !Ref ReadOnlyAccessArns
+                    - !Ref AWS::NoValue
+          Resource: "*"
```
zaro0508, line 295: wouldn't this give one user access to another's secrets?  might be better to add a prefix to your secrets so you can restrict access to the prefix? something like this.. `Resource: "/${AWS::StackName}/*`

templates/tower-project.yaml
```diff
@@ -0,0 +1,338 @@
...
+      SecretString: !Sub
+        - >-
+          {
+            "forge_access_key_id":      "${ForgeAccessKeyId}",
+            "forge_secret_access_key":  "${ForgeSecretAccessKey}",
+            "bucket_name":              "${BucketName}",
+            "read_write_arns":          "${ReadWriteArns}",
+            "read_only_arns":           "${ReadOnlyArns}",
+            "vpc_id":                   "${VpcId}",
+            "subnet_ids":               "${SubnetIds}"
```
zaro0508, line 245: these are not secrets, i usually find it to be a pain to keep non-secrets in a secrets store. is there a reason your storing them in a secrets store?

config/develop/example-project.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - common/nextflow-forge-iam-policy.yaml
+  - common/nextflow-launch-iam-policy.yaml
+
+parameters:
+  ReadWriteAccessArns:
+    - arn:aws:sts::563295687221:assumed-role/AWSReservedSSO_Developer_baa6fed639faf5e7/bruno.grande@sagebase.org
+  ReadOnlyAccessArns:
+    - arn:aws:sts::563295687221:assumed-role/AWSReservedSSO_Developer_baa6fed639faf5e7/tess.thyer@sagebase.org
+  AdminArns:
+    - {{stack_group_config.sso_admin_role.arn}}
```
zaro0508, line 14: i guess you can also do a lookup with `!rcmd` resolver if you want, although i'm not sure that's much better than passing in static string.. 

```
- !rcmd  aws iam get-role --role-name AWSReservedSSO_Administrator_5b839ce6e3a82407 --query 'Role.Arn'
```

zaro0508 (changes requested):
I'm glad we got away from those crazy joins and splits.  this version is much more readable.

templates/tower-project.yaml
```diff
@@ -0,0 +1,338 @@
...
+          PolicyDocument:
+            Version: "2012-10-17"
+            Statement:
+              Effect: Allow
+              Action:
+                - "s3:ListBucket"
+                - "s3:GetObject"
+                - "s3:PutObject"
+                - "s3:PutObjectTagging"
+                - "s3:DeleteObject"
```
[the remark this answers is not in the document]
BrunoGrandePhD: This list of actions was taken from [here](https://github.com/seqeralabs/nf-tower-aws/blob/master/launch/s3-bucket-write.json). It's what's suggested by Seqera, the company behind Nextflow Tower.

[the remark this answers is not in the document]
BrunoGrandePhD: But I'm running into S3 permission issues, so I might try broadening access.

templates/tower-project.yaml
BrunoGrandePhD: I wouldn't expect a resource-based policy bound to a specific secret to have any effect on other secrets. That said, I replaced the `Resource` to be `!Ref TowerProjectConfigurationSecret`.

templates/tower-project.yaml
BrunoGrandePhD: The only reason is because I already had a secret that users need to access, so I didn't want to create yet another resource with another ARN that I need to figure out, especially how that we've moved to cross-account access, which precludes access to CloudFormation outputs.

templates/tower-project.yaml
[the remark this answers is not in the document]
BrunoGrandePhD: I tried using `s3:*`, but that didn't resolve my issue, so I'm leaving this list as is for now.

config/develop/example-project.yaml
BrunoGrandePhD: In this case, this adds a command that can be skipped. That said, if ever we need any other information on the role, I like this approach since you gain access to everything returned by `get-role`.

templates/tower-project.yaml
```diff
@@ -0,0 +1,454 @@
...
+                - ecs.amazonaws.com
+                - batch.amazonaws.com
+            Action:
+              - sts:AssumeRole
+      Policies:
+        - PolicyName: "ForgeBatchJobPolicy"
+          PolicyDocument: 
+            Version: "2012-10-17"
+            Statement:
+              - Sid: ForgeInstanceRoleBatchJob
```
tthyer, line 138: This appears to largely be a duplication of an existing stack, one of the first created in the account,
`nextflow-forge-iam-policy`. I _generally_ find it preferable to use managed policies rather than inline ones, and you can making modifications to the existing one if necessary, since the original rollout of two required policies and a service user is being updated by this template.

tthyer (changes requested):

config/develop/example-project.yaml
```diff
@@ -0,0 +1,26 @@
...
+    - !stack_output_external workflows-nextflow-ci-service-account::ServiceRoleArn
+  NonAdminArns:
```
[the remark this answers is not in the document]
BrunoGrandePhD: Agreed. I prefixed the first two with `S3` and the last two with `Account`.

templates/tower-project.yaml
BrunoGrandePhD: Sorry, I pushed this right before logging off so that I could link to the template I was testing for Seqera. If this ends up being necessary, most of it will be moved to a managed policy. The only inline policy will be for resource-specific statements (like the KMS key).

templates/tower-project.yaml
```diff
@@ -0,0 +1,454 @@
...
+    Type: AWS::S3::BucketPolicy
+    Properties:
+      Bucket: !Ref TowerBucket
+      PolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+          - Sid: DenyAccessToNonAdmins
+            Effect: Deny
+            Principal:
+              AWS: !Ref NonAdminArns
```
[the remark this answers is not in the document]
BrunoGrandePhD: Because the SSO `Viewer` role is within the account and is bound to the AWS-managed `ViewOnlyAccess` policy, anyone assuming that role will have read-access to the project-specific buckets if we don't include this Deny statement (if I'm not mistaken). 

I would like to get rid of the `AccountNonAdminArns` (formerly `NonAdminArns`) if possible. We could add a Deny for `s3:*` on the SSO `Viewer` role and expect users to use their sandbox `Developer` role to access their project bucket. My only hesitation is whether blocking all S3 access on the `Viewer` role will make debugging more difficult. I can't think of anything, and we could pursue this approach now and change it later if we do find a fatal flaw.

templates/tower-project.yaml
BrunoGrandePhD: Another reason is that there's no way to attach a resource-based policy to an SSM Parameter for cross-account access. From what I can tell, this would require a role in our account that the external user/role would need to assume, which we just got rid of from this template.

templates/tower-project.yaml
BrunoGrandePhD: I updated the link I gave to Seqera to a commit permalin. I removed the InstanceRole from the template and disabled bucket encryption until Seqera offers us a workaround (so that I can test things in the meantime).

templates/tower-project.yaml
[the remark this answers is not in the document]
tthyer: As discussed in phone call, I prefer to revisit what permissions the Viewer role has rather than using a Deny. Let's opt-in to a narrower range of permissions.

Looking it up, the Viewer has AWS managed policy `ViewOnlyAccess` which includes `s3:ListAllMyBuckets` and `s3:ListBucket`. 
ListBucket allows seeing the objects within a bucket, but not pulling those.
The misleadingly named ListAllMyBuckets allows listing of all buckets, according this [knowledge center article](https://aws.amazon.com/premiumsupport/knowledge-center/s3-console-access-certain-bucket/).

Since ViewOnlyAccess allows access to many things that don't seem necessary, let's craft our own, narrower policy that does not allow our Viewer role to see things in S3. But we don't need to use a Deny. We can explicitly opt-in as @zaro0508 suggests.

templates/tower-project.yaml
```diff
@@ -0,0 +1,365 @@
...
+        Version: "2012-10-17"
+        Statement:
+          - Sid: DenyAccessToNonAdmins
+            Effect: Deny
+            Principal:
+              AWS: !Ref AccountNonAdminArns
+            Action: "s3:*"
+            Resource:
+              - !Sub "arn:aws:s3:::${TowerBucket}"
+              - !Sub "arn:aws:s3:::${TowerBucket}/*"
```
tthyer, line 183: Instead of Deny, let's rewrite which permissions the Viewer role has. 
This only matters for Prod account b/c PHI will only be there. So Developer role doesn't matter -- it's not in Prod.

tthyer (changes requested):

templates/tower-project.yaml
BrunoGrandePhD: The limited SSO Viewer role is implemented in Sage-Bionetworks-IT/organizations-infra#282.

templates/tower-project.yaml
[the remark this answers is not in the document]
BrunoGrandePhD: Done in Sage-Bionetworks-IT/organizations-infra#282

templates/tower-project.yaml
```diff
@@ -0,0 +1,354 @@
...
+            - HasS3ReadWriteAccessArns
+            - Sid: GrantReadAccessToS3ReadWriteAccessArns
+              Effect: Allow
+              Action: "secretsmanager:GetSecretValue"
+              Principal:
+                AWS: !Ref S3ReadWriteAccessArns
+              Resource: !Ref TowerProjectConfigurationSecret
+            - !Ref AWS::NoValue
+          - !If
+            - HasS3ReadWriteAccessArns
```
zaro0508, line 294: should this be `HasS3ReadOnlyAccessArns` instead?

zaro0508 (changes requested):

templates/tower-project.yaml
BrunoGrandePhD: Good catch!

zaro0508 (approved):
phew! much better than were we started from 😄

BrunoGrandePhD:
I deleted the CFN stacks that were deployed by the `common/nextflow-forge-service-user.yaml` config that I deleted.

PR #53, opened 2021-09-02
[WORKFLOWS-44] Add Managed Scaling

tthyer:
Adds a CapacityProvider to implement managed scaling. This allows the cluster to add an instance which will accommodate a new, updated task before spinning down the old one. It takes 5-6 minutes for this turnover to happen and then another ~15 minutes to shut down the second cluster instance. Possibly we could shorten the latter by adjusting TargetCapacity -- but this seems to be working nicely now.

### DIFF
--- a/templates/nextflow-ecs-cluster.yaml
+++ b/templates/nextflow-ecs-cluster.yaml
@@ -31,19 +31,23 @@ Parameters:
 
   AutoscalingGroupMaxSize:
     Type: Number
-    Description: >
-      Specifies the number of instances to launch and register to the cluster.
+    Description: (Optional) Limit on number of ECS cluster instances
       Defaults to 1.
+    Default: '2'
+
+  AutoscalingGroupDesiredCapacity:
+    Type: Number
+    Description: (Optional) Ideal number of ECS cluster instances
     Default: '1'
 
   RootEbsVolumeSize:
     Type: Number
-    Description: Optional - Specifies the Size in GBs of the root EBS volume
+    Description: (Optional) Specifies the Size in GBs of the root EBS volume
     Default: 30
 
   EbsVolumeType:
     Type: String
-    Description: Specifies the Type of (Amazon EBS) volume.
+    Description: (Optional) Specifies the Type of (Amazon EBS) volume.
     AllowedValues:
       - io1
       - gp2
@@ -55,7 +59,7 @@ Parameters:
 
   RootDeviceName:
     Type: String
-    Description: Optional - Specifies the device mapping for the root EBS volume.
+    Description: (Optional) Specifies the device mapping for the root EBS volume.
     Default: '/dev/xvda'
 
 Resources:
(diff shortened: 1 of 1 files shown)


templates/nextflow-ecs-cluster.yaml
```diff
@@ -140,7 +158,10 @@ Resources:
       LaunchConfigurationName: !Ref EcsLaunchConfiguration
       MinSize: '0'
       MaxSize: !Ref AutoscalingGroupMaxSize
-      DesiredCapacity: !Ref AutoscalingGroupMaxSize
+      DesiredCapacity: !Ref AutoscalingGroupDesiredCapacity
+      TerminationPolicies:
+        - OldestLaunchConfiguration
+        - OldestInstance
```
zaro0508, line 164: any reason why not just let AWS manage this with `Default` setting?

templates/nextflow-ecs-cluster.yaml
```diff
@@ -67,6 +71,20 @@ Resources:
...
+  EcsCapacityProvider:
+    Type: AWS::ECS::CapacityProvider
+    Properties:
+      AutoScalingGroupProvider:
+        AutoScalingGroupArn: !Ref EcsAutoScalingGroup
+        ManagedScaling:
+          Status: ENABLED
+          MinimumScalingStepSize: 1
+          MaximumScalingStepSize: 1
+          TargetCapacity: 90
```
zaro0508, line 86: The docs seem to indicate that if status is set to `enabled` then AWS auto scales for you.  I think that means that it will just ignore these scaling params?  or am i reading it wrong?..

```
The ManagedScaling property specifies the settings for the Auto
Scaling group capacity provider.

When managed scaling is enabled, Amazon ECS manages the scale-in
and scale-out actions of the Auto Scaling group. Amazon ECS manages a 
target tracking scaling policy using an Amazon ECS-managed CloudWatch 
metric with the specified targetCapacity value as the target value for the 
metric. For more information, see Using Managed Scaling in the Amazon 
Elastic Container Service Developer Guide.

If managed scaling is disabled, the user must manage the scaling of the 
Auto Scaling group. 
```

zaro0508 (approved):
looks good in general.  can approve as-is but just had a few questions.

templates/nextflow-ecs-cluster.yaml
tthyer: @zaro0508 Good question. It's the opposite of how you're reading it; in my experience with clusters there's always some metric or metrics used as the measure so that it knows _when_ to autoscale. Now, TargetCapacity is not a required value so I assume there's some default value they're using, but they don't specify what it is. I went with 90 because the aws-samples code I was looking at used that value; it works well enough so I left it at that.

Look at this [set of instructions](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/asg-capacity-providers-create-capacity-provider.html) for creating the capacity provider in the console -- there's more explanation at #9. You can also read the [deep dive blog post](https://aws.amazon.com/blogs/containers/deep-dive-on-amazon-ecs-cluster-auto-scaling/).

templates/nextflow-ecs-cluster.yaml
tthyer: Yes, I want to be explicit about how this is going to behave so that anyone supporting it knows what to expect. It's also a little different from the default behavior. You can look that up.

PR #54, opened 2021-09-02
Restrict KMS key access to project-specific roles and admins

BrunoGrandePhD:
Tower jobs couldn't access project buckets because they didn't have access to the KMS key used for bucket encryption. As a temporary workaround, I opened up access to the KMS key to the entire account. In this PR, I create two additional IAM roles with the requisite permissions to act as the Batch head and compute/work job roles, _i.e._ corresponding to the **Head Job Role** and **Compute Job Role** advanced options. This allows me to selectively grant KMS key access to project-specific roles instead of the entire account.


templates/tower-project.yaml
```diff
@@ -108,6 +108,69 @@ Resources:
...
+              - batch:DescribeJobs
+              - batch:RegisterJobDefinition
+              - batch:DescribeJobDefinitions
+              - ecs:DescribeContainerInstances
+              - ecs:DescribeTasks
+              - ec2:DescribeInstances
+              - ec2:DescribeInstanceAttribute
+              - ec2:DescribeInstanceTypes
+              - ec2:DescribeInstanceStatus
+            Resource: "*"
```
tthyer, line 164: Going to approve this PR, but it's too bad this has to be a star. Ideally the batch permissions should only apply to the batch created for this Tower compute environment. But this is created before that exists. This would be one advantage of manual setup over forge.

tthyer (approved):

PR #58, opened 2021-09-07
add in challenge project

thomasyu888:
This project will house challenge infrastructure testing workflow


config/prod/challenge-project.yaml
```diff
@@ -0,0 +1,48 @@
+# (REQUIRED) Step 0: Duplicate this file and rename using the updated stack name (see Step 1)
+
+# (REQUIRED) Step 1: Update the stack name below by replacing "example" with your project name
+stack_name: challenge-project
```
BrunoGrandePhD, line 4: **Nit (_i.e._ [optional](https://google.github.io/eng-practices/review/reviewer/standard.html#:~:text=Reviewers%20should%20always%20feel%20free,they%20could%20choose%20to%20ignore.)):** Do you have a specific challenge in mind? You might want to be more specific in the stack name so you can have different challenges with different users.

BrunoGrandePhD (approved):

config/prod/challenge-project.yaml
thomasyu888: I didn't actually.  I was just hoping to create an overall challenge nextflow project to test the evaluation of  submissions using nextflow.  Is it preferred to create a nextflow project on a per challenge basis?

BrunoGrandePhD: For Tower, I don't think it matters. I just think it might get tricky in terms of data governance and controlling data access. For example, if Challenge X and Challenge Y are both run using this project and someone is onboarded to help with Challenge Y, it won't be easy to specifically grant them access to the inputs/outputs for Challenge Y and not Challenge X, which are stored in the same bucket.

BrunoGrandePhD: I added this note in the README:

> Given that access is granted to the entire bucket, you might want to create more specific Tower projects that provide more granular access control.

----

> Do you envision people removing nextflow projects as they finish up?

Ideally, people delete projects when they're no longer needed. Otherwise, they'll have to keep paying for the S3 storage. We could institute a system where we ping people every 6 months after a year to check if they still want their project.

BrunoGrandePhD (approved):

PR #59, opened 2021-09-08
[WORKFLOWS-39] Parametrize IAM resources rather than hard-coding

BrunoGrandePhD:
Now that #50 is merged, we can use the same pattern elsewhere and close [WORKFLOWS-39](https://sagebionetworks.jira.com/browse/WORKFLOWS-39).

### DIFF
--- a/templates/nextflow-aurora-mysql.yaml
+++ b/templates/nextflow-aurora-mysql.yaml
@@ -84,14 +84,10 @@ Resources:
         AdminPrincipalArns: !Join
           - ','
           - - !Sub 'arn:aws:iam::${AWS::AccountId}:root'
-            - !ImportValue
-              'Fn::Sub': '${AWS::Region}-workflows-nextflow-ci-service-account-ServiceRoleArn'
             - !Join [",", !Ref AccountAdminArns]
         UserPrincipalArns: !Join
           - ','
-          - - !ImportValue
-              'Fn::Sub': '${AWS::Region}-workflows-nextflow-ci-service-account-ServiceRoleArn'
-            - !Join [",", !Ref AccountAdminArns]
+          - - !Join [",", !Ref AccountAdminArns]
             - !GetAtt BootstrapLambdaRole.Arn
 
   AuroraMasterSecret:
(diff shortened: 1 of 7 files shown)


templates/nextflow-aurora-mysql.yaml
```diff
@@ -87,12 +86,12 @@ Resources:
           - - !Sub 'arn:aws:iam::${AWS::AccountId}:root'
             - !ImportValue
               'Fn::Sub': '${AWS::Region}-workflows-nextflow-ci-service-account-ServiceRoleArn'
-            - !FindInMap [Accounts, !Ref AWS::AccountId, AdminRole]
+            - !Join [",", !Ref AccountAdminArns]
```
zaro0508, line 89: you can probably remove `'${AWS::Region}-workflows-nextflow-ci-service-account-ServiceRoleArn'`  since it looks like it's already being passed into `AccountAdminArns`

templates/nextflow-aurora-mysql.yaml
```diff
@@ -87,12 +86,12 @@ Resources:
...
         UserPrincipalArns: !Join
           - ','
           - - !ImportValue
               'Fn::Sub': '${AWS::Region}-workflows-nextflow-ci-service-account-ServiceRoleArn'
-            - !FindInMap [Accounts, !Ref AWS::AccountId, AdminRole]
+            - !Join [",", !Ref AccountAdminArns]
```
zaro0508, line 94: same here

templates/nextflow-aurora-mysql.yaml
```diff
@@ -87,12 +86,12 @@ Resources:
           - - !Sub 'arn:aws:iam::${AWS::AccountId}:root'
```
zaro0508, line 86: [optional] to further simplify, you could just pass in this ARN as well.  Then this whole thing just becomes a reference to the parameter, `!Ref  AccountAdminArns`.

zaro0508 (changes requested):

templates/nextflow-aurora-mysql.yaml
BrunoGrandePhD: Good catch! My eyes went straight for `!FindInMap`.

templates/nextflow-aurora-mysql.yaml
BrunoGrandePhD: Is there a way of obtaining the account ID in the Sceptre config (like `${AWS::AccountId}` in CFN)?

tthyer: I don't think so.

tthyer (approved):

PR #65, opened 2021-09-11
[WORKFLOWS-58] Test `configure-tower-project.py` script

BrunoGrandePhD:
⚠️   **Depends on #64**   ⚠️

(Work in progress)


thomasyu888:
@BrunoGrandePhD You can point this branch into the branch you're working on for #64.  when you merge #64 and delete the branch, it will automatically change the branch here. This way you can more accurately see the changes.

BrunoGrandePhD:
@thomasyu888: I didn't realize that the branch would be automatically updated. Thanks for the tip!

bin/test_configure_tower.py
```diff
@@ -0,0 +1,237 @@
+#!/usr/bin/env python3
+
+from copy import deepcopy
+import os
+
+import boto3
+from botocore.stub import Stubber
+import pytest
+
+import configure_tower
```
thomasyu888, line 10: Will the testing work when you run from a different directory?

```
pytest bin/test_configure_tower.py 
```

Since `configure_tower` isn't a package, will pytest try to import a python packge?

bin/test_configure_tower.py
```diff
@@ -0,0 +1,237 @@
...
+        return configure_tower.Projects(CONFIG_DIR)
+
+    def test__init__(self, mocker):
+        mocker.patch.object(configure_tower.Projects, "extract_users")
+        projects = configure_tower.Projects(CONFIG_DIR)
+        projects.extract_users.assert_called_once()
+
+    def test_list_projects(self, projects, mocker):
+        valid_names = self.valid_project_filenames
+        invalid_names = ["ignore.yaml" + "baz-project.txt"]
```
thomasyu888, line 94: Is there a reason why this isn't just `['ignore.yamlbaz-project.txt']`

bin/test_configure_tower.py
BrunoGrandePhD: I plan on reorganizing the files once #62 and #64 are merged. I just don't want to create a headache when I merge changes from `main` down the road. I'm not sure what this will look like yet. Do you have any suggestions?

bin/test_configure_tower.py
```diff
@@ -0,0 +1,237 @@
...
+        )
+
+    def test_extract_users_invalid(self, mocker):
+        for invalid_project in self.invalid_projects:
+            mocker.patch.object(
+                configure_tower.Projects,
+                "load_projects",
+                return_value=iter([invalid_project]),
+            )
+            with pytest.raises(AssertionError):
```
thomasyu888, line 171: I know you're not done with the PR, but there are missing invalid test cases here.

bin/test_configure_tower.py
BrunoGrandePhD: This is a typo. The `+` is supposed to be a comma.

bin/test_configure_tower.py
BrunoGrandePhD: Which ones should be added? How many is appropriate to test for?

bin/test_configure_tower.py
thomasyu888: Hmm, I'm actually unsure, I do feel like a lot of stuff here could eventually be superceded by a tower python client.  I think there's many solutions.

* The more complex one: utilize `importlib` package and import it after getting the true path
* The easier one, change the working dir in the test script

```
import os
directory = os.path.dirname(os.path.abspath(__file__))
os.chdir(directory)
```

bin/test_configure_tower.py
thomasyu888: Ideally, you test for all three assertions, since there are three different cases that an assertion error is thrown.  That being said this could be overkill here.

One definitely over the top thing you can do is create a function for the validation:

```
def _validate_config():
     # assertions here
```

You then test that the three asserts happen for `_validate_config` and when you're testing the `load_projects` function, all you have to test for is that you've called `_validate_config` at least once.  That also could reduce the amount of class attributes you have in your Test class.  Again, I don't think it's necessary here, just some food for thought.

bin/test_configure_tower.py
BrunoGrandePhD: I'll leave this conversation as unresolved as a reminder to tackle this once the other PRs are merged.

bin/test_configure_tower.py
BrunoGrandePhD: I like the idea of a `validate_config()` function, especially since it helps simplify the target code and the tests. I implemented it in #64 (ea3079ad6206c48941d5f550fd0c8b78ba0783c6) and updated the tests in this PR (29541fcc0c42f5748e787a4a3caefaae49469b93).

PR #67, opened 2021-10-01
[WORKFLOWS-65] Create VPC gateway endpoints for S3

BrunoGrandePhD:
This PR creates a VPC gateway endpoint for S3 in each AWS account to simultaneously eliminate most inter-AZ and NAT gateway traffic.

### DIFF
--- a/config/develop/nextflow-vpc-endpoint.yaml
+++ b/config/develop/nextflow-vpc-endpoint.yaml
@@ -1,0 +1,10 @@
+template_path: remote/gateway-vpc-endpoint.yaml
+stack_name: nextflow-vpc-endpoint
+dependencies:
+  - develop/nextflow-vpc.yaml
+parameters:
+  VpcName: "nextflow-vpc"
+  ServiceName: "com.amazonaws.us-east-1.s3"
+hooks:
+  before_launch:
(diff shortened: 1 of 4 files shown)


tthyer:
For reference, this is the [template](https://github.com/Sage-Bionetworks/aws-infra/blob/v0.3.2/templates/gateway-vpc-endpoint.yaml) being used.
@BrunoGrandePhD FWIW this LGTM, I'm relying on @zaro0508's advice that this is all we need.

tthyer (approved):
Approved but needs approval from IT too

config/develop/nextflow-vpc-endpoint.yaml
```diff
@@ -0,0 +1,10 @@
+template_path: remote/gateway-vpc-endpoint.yaml
+stack_name: nextflow-vpc-endpoint
```
zaro0508, line 2: i would probably add `s3` this stack name because you may want to add other endpoints between this vpc and another service.  how about something like `s3-nextflow-vpc-endpoint`?

zaro0508 (approved):
looks good, just had a suggestion on stack name.

config/develop/nextflow-vpc-endpoint.yaml
BrunoGrandePhD: Good idea!

tthyer (approved):

PR #73, opened 2021-11-04
[WORKFLOWS-67] Create more granular budget alerts

BrunoGrandePhD:
I've adapted the [budget template](https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/Billing/budgets.yaml) from aws-infra and added cost filters. This way, we can create more granular budget alerts. 

I propose three budgets: compute costs, storage (S3) costs, and everything else. The compute costs will spike, so the threshold on that is higher and will probably need to be adjusted once we have a better sense of compute needs. The storage costs will gradually increase as more projects are onboarded and more workflows are run. The "other" category should capture basic operating costs, which are generally stable. This would have caught the NAT gateway traffic that happened in September. I've set a purposely low threshold on this. 

The thresholds were estimated based on current costs in our nextflow-prod AWS account. We can tweak these later. 

We could create a daily budget report in a separate PR. I wanted to tackle this first so I can report back to AWS Support.

### DIFF
--- a/config/common/budget-compute.yaml
+++ b/config/common/budget-compute.yaml
@@ -1,0 +1,8 @@
+template_path: budget.yaml
+stack_name: budget-compute
+parameters:
+  budgetName: budget-compute
+  emailAddress: "{{stack_group_config.notifications_email}}"
+  budgetAmount: "5000"
+  serviceFilters:
+    {{stack_group_config.services.compute}}
--- a/config/common/budget-other.yaml
+++ b/config/common/budget-other.yaml
@@ -1,0 +1,8 @@
+template_path: budget.yaml
+stack_name: budget-other
+parameters:
+  budgetName: budget-other
+  emailAddress: "{{stack_group_config.notifications_email}}"
+  budgetAmount: "500"
+  serviceFilters:
+    {{stack_group_config.services.other}}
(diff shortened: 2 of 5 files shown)


thomasyu888:
LGTM. Will you have to continue adding to the "other" list or is this "other" list curated already to services commonly used at sage?

BrunoGrandePhD:
The "other" list is everything else that appeared in the Budgets console. Unfortunately, I cannot find a comprehensive list of services with the exact spelling that Budgets expects. Unfortunately, I had to scrape the current values from the "Create Budget" webpage (see docs for `serviceFilters` param for more details). 

So, these services are probably the ones that we've already encountered in our account. If we wanted to be even more comprehensive, I would need access to scrape the Budgets page for the AWS account that has the most diverse costs (presumably the account that is at the head of the AWS Organization).

tthyer:
LGTM

thomasyu888 (approved):

zaro0508 (approved):
looks good.  I'm just wondering if you need both this and cost anomaly detection in PR https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/310 ?  I think notifications might get a bit spammy?

PR #77, opened 2021-11-18
[WORKFLOWS-62 | WORKFLOWS-76] Add `CostCenter` tags

BrunoGrandePhD:
This PR comes after a discussion with IT on how we can best implement tagging in a way that's more useful for Finance. For more context, I suggest you read through [WORKFLOWS-76](https://sagebionetworks.jira.com/browse/WORKFLOWS-76) and [IT-796](https://sagebionetworks.jira.com/browse/IT-796). I believe these AWS accounts will be the first to implement this approach, so I'm all ears for feedback. 

You can skip to [this point](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/pull/77/files#diff-541004e9a888f44ff5dd5471545457a40141db89a4dc344a364bd83455324542) in the "Files changes" tab to see examples of program codes being used (everything else is just using a template variable). 

I'm also tackling [WORKFLOWS-62](https://sagebionetworks.jira.com/browse/WORKFLOWS-62) to reduce repetition of our default stack tags.

### DIFF
--- a/config/projects/amp-ad-project.yaml
+++ b/config/projects/amp-ad-project.yaml
@@ -22,4 +22,4 @@ stack_tags:
   Department: NDR
   Project: amp-ad
   OwnerEmail: william.poehlman@sagebase.org
-  SageFinanceProgramCode: NIA AMP-AD CC | 10015
+  CostCenter: NIA AMP-AD CC | 10015
--- a/config/projects/amp-ad-project.yaml
+++ b/config/projects/amp-ad-project.yaml
@@ -22,3 +22,4 @@ stack_tags:
   Department: NDR
   Project: amp-ad
   OwnerEmail: william.poehlman@sagebase.org
+  SageFinanceProgramCode: NIA AMP-AD CC | 10015
--- a/config/common/budget-compute.yaml
+++ b/config/common/budget-compute.yaml
@@ -1,8 +1,12 @@
 template_path: budget.yaml
 stack_name: budget-compute
+
 parameters:
   budgetName: budget-compute
   emailAddress: "{{stack_group_config.notifications_email}}"
   budgetAmount: "5000"
   serviceFilters:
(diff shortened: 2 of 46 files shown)


BrunoGrandePhD:
@brucehoff: I can't request a review from you yet since you're not a member of this GitHub organization (invite sent), but I'm tagging you in case you're interested.

thomasyu888 (approved):
Tremendous effort @BrunoGrandePhD .  LGTM

tthyer (approved):

config/projects/amp-ad-project.yaml
```diff
@@ -22,3 +22,4 @@ stack_tags:
   Department: NDR
   Project: amp-ad
   OwnerEmail: william.poehlman@sagebase.org
+  SageFinanceProgramCode: NIA AMP-AD CC | 10015
```
zaro0508, line 25: would be nice to make stack_tags resolvable (i.e. https://github.com/Sceptre/sceptre/pull/847), that way we could pull all these tags in from some type of spreadsheet.

also, not a big deal, the term AWS likes to use for this is `CostCenter` https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html

zaro0508 (approved):

config/projects/amp-ad-project.yaml
BrunoGrandePhD: @brucehoff: What do you think about `SageFinanceProgramCode` vs `CostCenter`?

brucehoff (commented):
The outstanding issue is that if the list of CostCenters defined by Sage changes then the tags added here will be wrong.  I would like to see a central list of CostCenters (that Sage can maintain) and a mechanism (like a Service Control Policy) to make sure that tags added to resources here conform to that list.

PR #78, opened 2021-11-18
Migrate important info to Confluence wiki page with watchers

BrunoGrandePhD:
In this PR, I propose using a Confluence wiki page to disseminate important information. I was initially going to create a Google Group, but I like how diffs are emailed to watchers on Confluence. Edits can optionally be published without notifying users in case we need to add less critical information. 

An added benefit of this approach is that we don't have to trigger our CI workflow with frequent edits to the README. 

Here's the Confluence page: https://sagebionetworks.jira.com/l/c/1hm4NP7a. Only Tess, Tom, and I have edit-access at the moment. Let me know if there's any important information that should be added.

### DIFF
--- a/.github/workflows/aws-deploy.yaml
+++ b/.github/workflows/aws-deploy.yaml
@@ -1,6 +1,9 @@
 name: aws-deploy
 
-on: push
+on:
+  push:
+    paths-ignore:
+      - '**.md'
 
 jobs:
   pre-commit:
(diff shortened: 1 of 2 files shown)


tthyer (approved):
I approve of moving anything that is Sage-related to Confluence. nextflow-infra could still be used by people outside Sage, so ideally the README would be for that audience.

Regarding trigger builds with README edits, there's a way to skip builds for those edits:
```
on:
  push:
    paths-ignore:
      - README.md
 ```

BrunoGrandePhD:
@tthyer: I should have known that there was a way to exclude files from the CI trigger. I added an exclusion for Markdown files. The updates to the important information are intended for Sage staff on how to use our internal Tower instance. The mechanics of our repository can remain documented in README/CONTRIBUTING (at least until we set up a library repository).

PR #79, opened 2021-11-18
Added comments inline

ahayden:
This PR Is not to me merged. It is to open discussion on the template only

### DIFF
--- a/templates/tower-project.yaml
+++ b/templates/tower-project.yaml
@@ -115,10 +115,10 @@ Resources:
             Action: sts:AssumeRole
             Principal:
               AWS: !GetAtt TowerForgeServiceUser.Arn
-          - Effect: Allow
+          - Effect: Allow  # Do account admins really need to be able to assume this role? for debugging?
             Action: sts:AssumeRole
             Principal:
-              AWS: !Ref AccountAdminArns
+              AWS: !Ref AccountAdminArns 
 
   TowerForgeBatchHeadJobRole:
     Type: AWS::IAM::Role
@@ -151,7 +151,7 @@ Resources:
   TowerForgeBatchHeadJobPolicy:
     Type: AWS::IAM::Policy
     Properties:
-      PolicyName: TowerForgeBucketAccessPolicy
+      PolicyName: TowerForgeBucketAccessPolicy #This policy allows more actions and should be renamed
       PolicyDocument:
         Version: '2012-10-17'
         Statement:
@@ -173,7 +173,7 @@ Resources:
               - ec2:DescribeInstanceAttribute
               - ec2:DescribeInstanceTypes
               - ec2:DescribeInstanceStatus
-            Resource: "*"
+            Resource: "*" #Should this be conditional on project resources only?
           - Sid: IAMConfig
             Effect: Allow
             Action:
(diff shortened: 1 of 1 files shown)


templates/tower-project.yaml
```diff
@@ -173,7 +173,7 @@ Resources:
               - ec2:DescribeInstanceAttribute
               - ec2:DescribeInstanceTypes
               - ec2:DescribeInstanceStatus
-            Resource: "*"
+            Resource: "*" #Should this be conditional on project resources only?
```
BrunoGrandePhD, line 176: Ideally, yes. Unfortunately, as Tess pointed out [here](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/pull/54#discussion_r701339679), we can't because the resources don't exist at the time of deployment. 😔

templates/tower-project.yaml
```diff
@@ -225,7 +225,7 @@ Resources:
   TowerBucket:
     Type: AWS::S3::Bucket
     Properties:
-      BucketName: !Sub "${AWS::StackName}-tower-bucket"
+      BucketName: !Sub "${AWS::StackName}-tower-bucket" #would it make sense to append project bucket names with a tenant name?
```
BrunoGrandePhD, line 228: These buckets might have multiple tenants (assuming you're referring to users who will be uploading and downloading to the bucket). Are you thinking of something else?

templates/tower-project.yaml
```diff
@@ -186,7 +186,7 @@ Resources:
   EncryptionKeyStack:
     Type: AWS::CloudFormation::Stack
     Properties:
-      TemplateURL: !Sub ${TemplateRootUrl}/aws-infra/v0.3.1/KMS/kms-key.yaml
+      TemplateURL: !Sub ${TemplateRootUrl}/KMS/kms-key.yaml
```
BrunoGrandePhD, line 189: @ahayden: I see the need to use the latest template for security reasons. However, I am afraid of deploying an unrelated change and things start breaking because the KMS key template works differently after an update. How should we balance these two goals?

BrunoGrandePhD (changes requested):
@ahayden: Thanks for the review!

BrunoGrandePhD:
@ahayden: Do you have any responses to my comments?

templates/tower-project.yaml
ahayden: What about if the condition statement is on [resource tags](https://docs.aws.amazon.com/IAM/latest/UserGuide/tutorial_attribute-based-access-control.html)? The resources don't need to exist, just need be tagged correctly when they are created.

templates/tower-project.yaml
ahayden: This is really just a style choice... My comment was meant so that you would pass in the library version with the `TemplateRootUrl` like 'https://remoterepo.tld/aws-infra/v0.3.1/'... and so would only have to change the parameter when if you wanted to inc library versions

templates/tower-project.yaml
ahayden: I was just thinking of the naming convention getting out of hand... if everything has a static prefix it would be simpler to find them later.

ahayden (commented):
Just commenting.. the pr is not to be merged

PR #83, opened 2021-12-08
[WORKFLOWS-84] Provide default GitHub credentials

BrunoGrandePhD:
Unfortunately, Nextflow Tower seems to issue plenty of requests to GitHub, which results in rate limits being imposed. These rate limits manifest in the form of uninformative errors like `Unknown pipeline repository or missing credentials`. In this PR, we are configuring Tower to authenticate requests to GitHub using a service account, which should have higher rate limits than unauthenticated requests. 

At the same time, I'm taking the opportunity to switch to using the `S3Objects` macro to obviate the need for a post-launch Sceptre hook. Shout-out to @tthyer for fixing the `S3Objects` macro!


tthyer (approved):
looks good

PR #84, opened 2021-12-09
[WORKFLOWS-85] Prototype Tower project in separate account

BrunoGrandePhD:
While I'm waiting for the STRIDES AMP-AD account to be created, I wanted to prototype the deployment of a Tower project to a separate account. In this PR, I'm using sandbox, but the intent is to switch to the new AMP-AD account once it's available. The benefit is that all compute and storage costs would get charged to NIH directly. 

Note that most of the "file changes" are files being renamed because I streamlined the GitHub Actions workflow. I've adopted what @tthyer did for BridgeDownstream, and to make it extra streamlined, I'm using prefixes and suffixes for the Sceptre stack groups. 

Depends on Sage-Bionetworks-IT/organizations-infra#327


config/infra-sandbox/nextflow-vpc.yaml
```diff
@@ -0,0 +1,11 @@
...
+  type: http
+  url: "{{stack_group_config.aws_infra_templates_root_url}}/v0.3.0/templates/VPC/vpc-20-private.yaml"
+stack_name: nextflow-vpc
+
+parameters:
+  VpcName: nextflow-vpc
+  VpcSubnetPrefix: "172.23"
+
+stack_tags:
+  {{stack_group_config.default_stack_tags}}
```
tthyer, line 11: Is there a reason you can't use sandcastle? Does it have to do with subnets perhaps?

BrunoGrandePhD: I considered it, but the Tower configuration script looks for a CFN stack called `nextflow-vpc` and currently assumes four private subnets (sandcastle has 3 public and 3 private). 

I figured that this VPC would eventually be re-deployed to the AMP-AD, so this PR can basically undergo `s/sandbox/ampad/g` and it should work.

.github/workflows/aws-deploy.yaml
```diff
@@ -55,61 +69,31 @@ jobs:
...
-          responseCode: 200
-          timeout: 180000
-          interval: 30000
-
       - name: Configure projects in Tower
-        run: pipenv run bin/configure-tower-projects.py config/projects/
+        run: pipenv run bin/configure-tower-projects.py config/projects-${{ matrix.sceptre-suffix }}
         env:
-          NXF_TOWER_TOKEN: ${{ secrets.TOWER_PROD_TOKEN }}
+          NXF_TOWER_TOKEN: ${{ secrets.TOWER_DEV_TOKEN }}
```
BrunoGrandePhD, line 98: I'll need to update this so the right token is used depending on the Tower URL.

BrunoGrandePhD: I moved the Tower tokens to the environments

PR #85, opened 2021-12-20
Increase max CPU count and head job memory

BrunoGrandePhD:
More details to come later.


BrunoGrandePhD:
I'll be making this change in a separate PR

PR #88, opened 2022-02-17
Fix dependency paths after re-organizing configs

BrunoGrandePhD:
I forgot to update the dependency paths in the Sceptre config files, so I ran into errors during deployment. 

https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/runs/5236025506?check_suite_focus=true


thomasyu888 (approved):

tthyer (approved):
I generally see what you're doing -- since you're adding some complexity and diverging from what's typically done for your infra repos that use sceptre, please make sure to update the Configuration section in CONTRIBUTING.md

BrunoGrandePhD:
@tthyer: Thanks for the prompt! The contribution guidelines definitely needed to be spruced up a little.

PR #93, opened 2022-02-26
Fix `CostCenter` tags values based on latest list

BrunoGrandePhD:


config/config.yaml
```diff
@@ -7,7 +7,7 @@ default_stack_tags:
   Department: IBC
   Project: Infrastructure
   OwnerEmail: nextflow-admins@sagebase.org
-  CostCenter: Indirects
+  CostCenter: NO PROGRAM / 000000
```
thomasyu888, line 10: Switched to backslash?

thomasyu888 (approved):
LGTM. Small comment

config/config.yaml
BrunoGrandePhD: Yeah, because the pipe isn't a supported character in tag values. Khai ran into this issue in the following PR after I had already updated nextflow-infra (but hadn't run into any deployment issues). 

https://github.com/Sage-Bionetworks/aws-infra/pull/329#issue-1083464281

PR #95, opened 2022-03-03
Create nfri-ctf-nf1-project.yaml

allaway:
Ok, I'm pretty sure this is my last project config file for a while... :) 

Thanks!

**Edit:**
Depends on #98


BrunoGrandePhD (approved):
There's a bug in the Tower configuration script, which is what creates the workspace in Tower and grants individuals with access. I would like to address that first before merging this to avoid messing things up.

allaway:
Makes sense, thanks!

PR #96, opened 2022-03-10
[WORKFLOWS-181] Create and add teams to Tower workspaces

BrunoGrandePhD:
We found that Tower users weren't getting access to the right workspaces. As a first-pass attempt to resolving this issue, I'm updating the Tower configuration script to create teams per project per role and adding these teams to the relevant workspaces. I'm also performing the additional step of removing unexpected users from each team to make sure they remain in sync with the Sceptre config file. 

I was able to test the updated script in Tower-dev. Once it's deployed to Tower-prod, I'm going to remove all of the users that are listed individually on workspaces. Because these users will have been added via a team, no one should suffer a lapse of access (unless they weren't supposed to have access to the workspace in the first place). 

**Edit:** As I show in [this comment](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/pull/96#issuecomment-1064220965), workspace participants with the `maintainer` role can add people to the workspace. I don't know if we can revoke this permission without affecting their ability to launch workflows. At the moment, the script ensures that the team membership is in sync with the Sceptre config file. Should we also make sure that no one else is added to the workspace as individual users (_i.e._ outside of a team)? 

## Organization Teams 

image

## Example Team

image

## Teams Added to Workspace

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Before you can use Nextflow Tower, you need to first deploy a Tower project, whi
 
    - A new Tower workspace called `<stack_name>` was created under this organization.
 
-   - Users listed under `S3ReadWriteAccessArns` were added as workspace participants with the `Maintain` role, which grants the following permissions:
+   - Users listed under `S3ReadWriteAccessArns` were added to a workspace team with the `Maintain` role, which grants the following permissions:
 
      > The users can launch pipeline and modify pipeline executions
        (e.g. can change the pipeline launch compute env, parameters,
@@ -63,7 +63,7 @@ Before you can use Nextflow Tower, you need to first deploy a Tower project, whi
        configuration in the Launchpad. The users cannot modify Compute
        env settings and Credentials
 
-   - Users listed under `S3ReadOnlyAccessArns` were added as workspace participants with the `View` role, which grants the following permissions:
+   - Users listed under `S3ReadOnlyAccessArns` were added to a workspace team with the `View` role, which grants the following permissions:
 
      > The users can access to the team resources in read-only mode
 
--- a/bin/configure-tower-projects.py
+++ b/bin/configure-tower-projects.py
@@ -36,8 +36,7 @@ def main() -> None:
         )
     else:
         tower = TowerClient()
-        org = TowerOrganization(tower, projects)
-        org.create_workspaces()
+        TowerOrganization(tower, projects)
 
 
 class InvalidTowerProject(Exception):
@@ -383,12 +382,12 @@ class TowerWorkspace:
         return response["workspace"]
 
     def add_participant(self, role: str, user: str = None, team_id: int = None) -> int:
-        """Add user to the workspace (if need be) and return participant ID
+        """Add user or team to the workspace (if need be) and return participant ID
 
         Args:
             role (str): 'owner', 'admin', 'maintain', 'launch', or 'view'
-            user (str): Email address for the user
-            team_id (int): Team identifier
+            user (str): Email address for the user. Mutually exclusive with `team_id`.
+            team_id (int): Team identifier. Mutually exclusive with `user`.
 
         Returns:
             int: Participant ID for the user in the given workspace
@@ -617,6 +616,7 @@ class TowerOrganization:
         self.members: Dict[str, dict] = dict()
         self.populate()
         self.workspaces: Dict[str, TowerWorkspace] = dict()
+        self.create_workspaces()
 
     def create(self) -> dict:
         """Get or create Tower organization with the given name
@@ -745,9 +745,7 @@ class TowerOrganization:
         """
         endpoint = f"/orgs/{self.id}/teams/{team_id}/members"
         response = self.tower.request("GET", endpoint)
-        team_member_ids = list()
-        for team_member in response["members"]:
-            team_member_ids.append(team_member["memberId"])
+        team_member_ids = [member["memberId"] for member in response["members"]]
         return team_member_ids
 
     def populate(self) -> None:
--- a/bin/configure-tower-projects.py
+++ b/bin/configure-tower-projects.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 import argparse
+from collections import defaultdict
 import json
 import os
 import re
@@ -88,12 +89,13 @@ class Users:
         self.launchers = launchers
         self.viewers = viewers
 
-    def list_users(self) -> Iterator[Tuple[str, str]]:
+    def list_users(self) -> Iterator[Tuple[str, str, str]]:
         """List all users and their Tower roles
 
         Yields:
-            Iterator[Tuple[str, str]]:
-                Each element is the user email (str) and Tower role (str)
+            Iterator[Tuple[str, str, str]]:
+                Each element is the user email (str), the user group,
+                and Tower role (str)
         """
         role_mapping = {
             "owners": "owner",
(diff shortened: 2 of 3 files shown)


bin/configure-tower-projects.py
```diff
@@ -365,37 +382,54 @@ def create(self) -> dict:
         response = self.tower.request("POST", endpoint, json=data)
         return response["workspace"]
 
-    def add_participant(self, user: str, role: str) -> int:
+    def add_participant(self, role: str, user: str = None, team_id: int = None) -> int:
         """Add user to the workspace (if need be) and return participant ID
```
thomasyu888, line 386: Nit: This function now adds both participant and team? Ok with function name but improve docs?

bin/configure-tower-projects.py
```diff
@@ -365,37 +382,54 @@ def create(self) -> dict:
...
+        elif not user and team_id:
+            identifier = team_id
+            data = {
+                "memberId": None,
+                "teamId": team_id,
+                "userNameOrEmail": None,
+            }
+        else:
+            raise ValueError(
+                "Must provide value for exactly one of `user` or `team_id`."
```
thomasyu888, line 415: nit: Why not user_id when its team_id?

bin/configure-tower-projects.py
```diff
@@ -640,15 +679,101 @@ def add_member(self, user: str) -> dict:
...
+        """Create team under organization with the given name
+
+        Args:
+            team_name (str): Team name
+
+        Returns:
+            int: Team identifier
+        """
+        # Check if the team already exists
+        endpoint = f"/orgs/{self.id}/teams"
```
thomasyu888, line 727: Nit: I saw this `/orgs/{orgId}/teams/validate`: Check that is a valid team name. 

Does this mean check if team exists? Or if the actual team name is compliant with nextflow tower.

bin/configure-tower-projects.py
```diff
@@ -640,15 +679,101 @@ def add_member(self, user: str) -> dict:
...
+            team_id (int): Team identifier
+
+        Returns:
+            List[int]: List of team member IDs
+        """
+        endpoint = f"/orgs/{self.id}/teams/{team_id}/members"
+        response = self.tower.request("GET", endpoint)
+        team_member_ids = list()
+        for team_member in response["members"]:
+            team_member_ids.append(team_member["memberId"])
```
thomasyu888, line 750: Nit: use list comprehension
```suggestion
        team_member_ids = [
            team_member["memberId"]
            for team_member in response["members"]
        ]
```

bin/configure-tower-projects.py
```diff
@@ -667,8 +792,9 @@ def create_workspaces(self) -> Dict[str, TowerWorkspace]:
             Dict[str, TowerWorkspace]:
                 Mapping of project names and their corresponding workspaces
         """
-        for name, users in self.list_projects():
-            ws = TowerWorkspace(self, name, users)
+        for name, _ in self.list_projects():
+            teams = self.teamids_per_project[name]
```
thomasyu888, line 796: Nit: It is more apparent after reading `__init__` that on the instantiation of the TowerOrganization class that both `self.create()` and `self.populate()` is called?  I can't really think of a scenario when `create_workspaces` would be called without the instantiation of the class.

bin/configure-tower-projects.py
```diff
@@ -640,15 +679,101 @@ def add_member(self, user: str) -> dict:
...
     def populate(self) -> None:
         """Add all emails from across all projects to the organization
 
         Returns:
             Dict[str, dict]: Same as self.project, but with member IDs
         """
-        for project_users in self.users_per_project.values():
-            for user, _ in project_users.list_users():
-                self.add_member(user)
+        for project_name, project_users in self.users_per_project.items():
```
thomasyu888, line 759: Apologies, as Im not entirely familiar, but where are these users for projects specified?

README.md
```diff
@@ -55,15 +55,15 @@ Before you can use Nextflow Tower, you need to first deploy a Tower project, whi
 
    - A new Tower workspace called `<stack_name>` was created under this organization.
 
-   - Users listed under `S3ReadWriteAccessArns` were added as workspace participants with the `Maintain` role, which grants the following permissions:
+   - Users listed under `S3ReadWriteAccessArns` were added to a workspace team with the `Maintain` role, which grants the following permissions:
```
thomasyu888, line 58: Are we the only admins?

thomasyu888 (commented):
Clever approach - I like how you make sure that the config file is the source of truth.  That said - are members of teams allowed to add users to teams in the UI?  Does it become an issue if people are adding members to teams and this stack is re-run.

BrunoGrandePhD:
@thomasyu888: The `nextflow-tower-admin` service account has the `owner` role on the Sage organization and can modify team membership. Everyone else just has the `member` role, which has limited read-only access to the organization. Here's what I see on the organization page with my own account:

image

Contrast that with what the owner sees:

image

That said, workspace maintainers (_i.e._ the users listed under `S3ReadWriteAccessArns`) can add users to a workspace. When this is done, the user is added with the `launch` role, which is only one step above the `view` role. Also, being added to the workspace doesn't grant you access to the bucket, so it's of limited usefulness and risk. 

image

A legitimate question is whether we also want to clean up the workspace participants to make sure everything stay in sync with the Sceptre config file, which would be the ground truth. I'm leaning towards yes, but I'm open to thoughts.

README.md
BrunoGrandePhD: The `nextflow-tower-admin` service account is the only owner on the organization, teams, and workspaces. This way, it's harder for the Tower setup to drift from what this script deploys.

bin/configure-tower-projects.py
BrunoGrandePhD: It does alert you when the team name already exists, and it also checks if the team name contains valid characters. The former case is handled with the existing code, and the latter case shouldn't happen because we're appending `-<role>` to the stack name, which already only consists of valid characters. 

I could use the dedicated `/validate` endpoint to check if a team name already exists. Can you think of any benefits? 

```
>>> tower.request("GET", "/orgs/193738885884191/teams/validate?name=example-dev-project-maintainers")
{'message': "A team with name 'example-dev-project-maintainers' already exists"}

>>> tower.request("GET", "/orgs/193738885884191/teams/validate?name=#winning")
{'message': 'Team name must contain a minimum of 1 and a maximum of 38 alphanumeric characters separated by dashes or underscores'}
```

bin/configure-tower-projects.py
BrunoGrandePhD: If something were to be `user_id`, it would be the `member_id`. Currently, `user` contains emails because that's what's stored in the Sceptre config files. I'm not using `email` as the variable name because it can technically contain a username (as indicated by the Tower field, `userNameOrEmail`). I'm not particularly eager to use `user_name_or_email` either. 

Can you think of a better variable name?

BrunoGrandePhD: I could refactor the code so that member IDs are tracked as soon as they're obtained instead of relying on `self.org.members` to retrieve the member ID in this function.

bin/configure-tower-projects.py
BrunoGrandePhD: This isn't a nit. The docs were inaccurate, so I've updated them. "Participant" is the term that Tower uses to refer to users or teams added to a workspace.

bin/configure-tower-projects.py
thomasyu888: I see. This makes sense. No need to refactor, user makes sense here.

bin/configure-tower-projects.py
thomasyu888: Are the teams scoped to the tower project? The efficiency would be not having to loop through a list of teams as the team list grows but it's really not too big of a burden

bin/configure-tower-projects.py
thomasyu888: Agreed, the docstring for this function still seems unchanged.

BrunoGrandePhD: Sorry, I forgot to upload my commit.

thomasyu888 (approved):
LGTM : up to you if you want to make the change with the validation of the team name.  Even with hundreds of teams, it won't be computationally expensive.

bin/configure-tower-projects.py
BrunoGrandePhD: > Are the teams scoped to the tower project?

There is one team per project per role (_i.e._ read/write aka maintainer and read-only aka viewer). These teams exist under the organization and aren't tied to the project workspace until they're added to the workspace as participants with the relevant role. I'm not sure if this answers your question. 

We'll need to iterate through the teams anyways to make sure their membership is up-to-date. We can see what the runtime is once we deploy to production. The dev environment only has two teams, so it's hard to tell.

PR #97, opened 2022-03-16
[WORKFLOWS-181] Ensure shorter team names due to length limit

BrunoGrandePhD:
I ran into the following error after merging #96 when creating the `challenge-infra-test-project-maintainers` team in Tower-prod. I've added a note to the README, and I'm trimming the `-project` suffix from the team name to compensate for the addition of the `-maintainers` or `-viewers` suffix. 

```
{'message': 'Team name must contain a minimum of 1 and a maximum of 38 alphanumeric characters separated by dashes or underscores'}
```

I'm disabling debug mode now that I figured out the issue with the #96 deployment.

PR #98, opened 2022-03-17
[WORKFLOWS-181] Add support for paged requests

BrunoGrandePhD:
After merging #96 and #97, I realized that people were still missing from certain workspaces. I finally identified the root cause of WORKFLOWS-181. Because there are more than 10 members in our Sage Tower organization, responses started to span multiple pages, but my Tower client was only considering the first page of results (10). 

Sadly, this wasn't caught because I had a [poorly written for-loop](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/f6f0de41f010195ed9a46979936ff4ab89a4b2cf/bin/configure-tower-projects.py#L683-L685) that was defaulting to the last member in the list. It's a silly mistake on my part. I've addressed both instances of this pattern with a more defensive approach. 

In reviewing the [Tower API docs](https://tower.nf/openapi/index.html) for how to page results, I came across the `search` query parameter, which exists for some endpoints. I've deployed it where applicable.

### DIFF
--- a/bin/configure-tower-projects.py
+++ b/bin/configure-tower-projects.py
@@ -328,21 +328,41 @@ class TowerClient:
         assert method in {"GET", "PUT", "POST", "DELETE"}
         url = self.tower_api_base_url + endpoint
         kwargs["headers"] = {"Authorization": f"Bearer {self.tower_token}"}
-        if self.debug:
-            print(f"\nEndpoint:\t {method} {url}")
-        if self.debug and "params" in kwargs:
-            print(f"Params: \t {kwargs['params']}")
-        if self.debug and "json" in kwargs:
-            print(f"Payload:\t {kwargs['json']}")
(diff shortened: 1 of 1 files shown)


bin/configure-tower-projects.py
```diff
@@ -328,22 +328,42 @@ def request(self, method: str, endpoint: str, **kwargs) -> dict:
...
+
+        Args:
+            method (str): An HTTP method (GET, PUT, POST, or DELETE)
+            endpoint (str): The API endpoint with the path parameters filled in
+
+        Returns:
+            Iterator[Dict]: An iterator traversing through pages of responses
+        """
+        params = kwargs.pop("params", {})
+        params["max"] = 10
```
BrunoGrandePhD, line 355: Consider using 100

BrunoGrandePhD:
Performed an in-person code review

PR #99, opened 2022-03-27
Ensure that database migration completes before launching other containers

BrunoGrandePhD:
For WORKFLOWS-155, I'll need to update the Tower containers. According to the [docs](https://git.seqera.io/sage/nf-tower-deployment/src/branch/master/release-notes-21.12.md), the `/migrate-db.sh` script should be run before anything else. It's indeed run first in the `cron` container, but I don't know whether the dependencies that `backend` and `frontend` have on `cron` ensure that they will only start once the `migrate-db.sh` script is finished running. In other words, I don't know exactly when the `START` condition of the [container dependencies](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-containerdependency.html) is met. 

Essentially, I split off the execution of the `migrate-db.sh` script in a separate container, which is expected to terminate successfully. Only then will the downstream `cron`, `backend`, and `frontend` containers start. 

I realize this introduces code duplication, so I'm eager to hear feedback on the tradeoffs.

PR #100, opened 2022-03-27
[WORKFLOWS-155] Enable secrets and OIDC token import in Tower-Dev

BrunoGrandePhD:
I'm deploying a beta version of the Tower containers so I can test the OIDC login for Paolo. This change is going to be restricted to the development environment until a stable release is minted.

Depends on #99


thomasyu888 (approved):
LGTM

PR #103, opened 2022-03-31
[WORKFLOWS-187] Add scratch buckets for each Tower project

BrunoGrandePhD:
This PR will add a separate bucket with a default data expiration of six months. It'll be configured in #101 as the default work directory for the `v3` compute environments. I will also update the internal docs so they recommend to use this scratch bucket for staging Synapse files with `synstage`. 

Shout out to @tthyer for the elegant idea!


thomasyu888 (approved):
One comment - can you explain to me why SynapseIndexing is disabled?

PR #104, opened 2022-03-31
[WORKFLOWS-210] Configure Synapse OIDC as an authentication option

BrunoGrandePhD:
Depends on #101 

In preparation to using the new Tower secrets functionality and the capture of the OIDC refresh JWT, I'm adding Synapse as an identity provider. I still need to validate the clients before they can work, and I'll look into how to configure the allow-list for Synapse (not sure if it filters on Synapse usernames or emails). 

I won't merge this until we confirm the behavior in Tower-dev. 

https://tower-dev.sagebionetworks.org/

image


src/tower/resources/environment.yaml
```diff
@@ -15,5 +15,8 @@ TOWER_LICENSE: "!Ref 'TowerLicense'"
 FLYWAY_LOCATIONS: "classpath:db-schema/mysql"
 TOWER_GOOGLE_CLIENT: "!Ref TowerGoogleClientId"
 TOWER_GOOGLE_SECRET: "!Ref TowerGoogleSecret"
+TOWER_OIDC_CLIENT: "!If [!Ref HasTowerOidcClient, !Ref TowerOidcClientId, !Ref AWS::NoValue]"
+TOWER_OIDC_SECRET: "!If [!Ref HasTowerOidcClient, !Ref TowerOidcSecret, !Ref AWS::NoValue]"
+TOWER_OIDC_ISSUER: "!If [!Ref HasTowerOidcClient, !Ref TowerOidcIssuer, !Ref AWS::NoValue]"
```
BrunoGrandePhD, line 20: @zaro0508: I'm trying to set these environment variables only when the `HasTowerOidcClient` condition evaluates to true. I don't think setting the `Value` of an environment variable to `AWS::NoValue` will have the intended effect.

BrunoGrandePhD: Never mind. I found a simpler way. Let me push that commit.

thomasyu888 (approved):

PR #107, opened 2022-04-06
[IT-1781] Clean up AWS resources (security tweaks)

BrunoGrandePhD:
The goal of this PR is to perform a few security tweaks:

1. Upgrade the SSL/TLS policy of the Tower load balancer to something that meets Sage IT's standards
    * I believe no further edits are required to meet this request. 
2. Deploy maintenance tasks for the Tower EC2 instances
    * I have added the `essentials` and `maintenance` stacks
    * However, I believe that I need to bind the `ManagedInstanceRole` instance profile to the EC2 instances provisioned by ECS
    * I'll need to do a bit of digging to figure out how to do this

In the meantime, feel free to comment with suggestions!


config/common/essentials.yaml
```diff
@@ -0,0 +1,13 @@
+template:
+  type: http
+  url: "{{stack_group_config.aws_infra_templates_root_url}}/v0.3.8/templates/essentials.yaml"
+stack_name: essentials
+
+parameters:
+  VpcPeeringRequesterAwsAccountId: '745159704268'
```
BrunoGrandePhD, line 7: Not sure if this is relevant anymore with the new TGW setup, but it's a required parameter, so I don't have a choice.

zaro0508: newer accounts (created with org-formation) should need essentials.yaml.  it creates a bunch of resources that are not really needed anymore.  If maintenance depends on essential resources then IT needs to fix that.

zaro0508: looking at the [maintenance template](https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/maintenance.yaml), I don't see any dependencies on resources from essentials.yaml.  Maybe all you need to do is drop essentials.yaml from this PR.

thomasyu888 (approved):

ahayden (approved):

PR #109, opened 2022-04-13
[WORKFLOWS-233] Upgrade Nextflow Tower to `v22.1.0`

BrunoGrandePhD:
This PR deploys the latest version of Nextflow Tower. One of the main new features is secrets management. I was able to run `synstage` using a `SYNAPSE_AUTH_TOKEN` secret in Tower-dev. Another useful feature is the admin panel, which can be enabled for a select few root users. 

This update also required tweaking a few IAM roles based on the instructions [here](https://github.com/seqeralabs/nf-tower-aws/tree/master/launch). This was necessary for enabling the secrets functionality. 

Before merging, I'm going to have to delete unused compute environments to avoid running into the limit of 50 Batch CEs.


thomasyu888 (approved):
So, all the batch compute environments need to be recreated and all the older compute environments should be removed right?

bin/configure-tower-projects.py
```diff
@@ -556,6 +556,8 @@ def cleanup_compute_environments(self):
         for comp_env in response["computeEnvs"]:
             comp_env_id = comp_env["id"]
             comp_env_name = comp_env["name"]
+            if comp_env_name.endswith(CE_VERSION):
```
thomasyu888, line 559: What happens if a compute environment is in use?  Can it not be deleted?

BrunoGrandePhD: It cannot, and I'm relying on this fact since it would be complicated to replicate this behavior myself. 

For compute environments with running jobs, the `DELETE` request fails with:
```
 ERROR: Compute environment '1GZDXqNBbdeWFRdaWnGbKE' has active jobs
```

PR #111, opened 2022-04-26
Disable private Tower workspaces per user

BrunoGrandePhD:
Disabling the user workspaces, which cannot be used anyways since they don't have credentials, will reduce user confusion.


thomasyu888 (approved):
LGTM

PR #112, opened 2022-04-26
Consolidate and update Tower version

BrunoGrandePhD:
A patch release was just published by Seqera, so I'm updating Tower to benefit from any bug fixes.


thomasyu888 (approved):
Good idea to add to the config.yaml

PR #114, opened 2022-05-11
[WORKFLOWS-258] Block all public access on S3 buckets

BrunoGrandePhD:
CloudFormation defaults to **not** blocking public access on S3 buckets. This PR addresses that for public that ought not to be public.


thomasyu888 (approved):
LGTM!

PR #115, opened 2022-06-07
Refine budget limits per account

BrunoGrandePhD:
We've increased the amount of activity in nextflow-prod, so the budget limits needed to be tweaked. I've also noticed that GuardDuty and AWS Config costs scale roughly with compute, so I categorized them separately to keep better track of other costs. In this PR, I move the budget limits to the config files and tailor them per account.


config/infra-dev/config.yaml
```diff
@@ -1,2 +1,5 @@
 sso_admin_role:
   arn: arn:aws:iam::035458030717:role/aws-reserved/sso.amazonaws.com/AWSReservedSSO_Administrator_580e9f32ac55c4e7
+budgets:
+  compute: 1000
```
thomasyu888, line 4: Just as a reminder for myself, these budgets override what's in `config/config.yaml` correct?

thomasyu888 (approved):
LGTM

config/infra-dev/config.yaml
BrunoGrandePhD: Yup! https://docs.sceptre-project.org/dev/docs/stack_group_config.html#stack-group-config-cascading-config

config/infra-ampad/config.yaml
```diff
@@ -2,3 +2,6 @@
 # it sits outside of our AWS Organization fo STRIDES billing
 sso_admin_role:
   arn: arn:aws:iam::751556145034:role/strides-ampad-workflows-admin
+budgets:
+  compute: 5000
+  storage: 2000
```
zaro0508, line 7: the same values are in `config/config.yaml` so do we need this here?

src/sceptre/variables/ampad.yaml
```diff
@@ -0,0 +1,4 @@
+compute_budget: 5000
+storage_budget: 2000
+security_budget: 500
+other_budget: 500
```
zaro0508, line 4: make strings instead of numbers

zaro0508 (changes requested):

zaro0508 (approved):

PR #117, opened 2022-06-13
Refine budgets again based on notifications

BrunoGrandePhD:
The current budget thresholds are a little noisy. I'm replacing an 80% threshold with a 200% threshold so that I can keep tabs on an expected surge in cost past 100%. I'm also moving EBS from `other` to `storage` since that cost will roughly correlate with compute. 

**Note to self:** Based on my test in nextflow-dev, I'll need to manually delete the budget CFN stacks before deploying this.

### DIFF
--- a/templates/budget.yaml
+++ b/templates/budget.yaml
@@ -132,14 +132,6 @@ Resources:
           IncludeTax: false
       NotificationsWithSubscribers:
         - Notification:
-            NotificationType: ACTUAL
-            ComparisonOperator: GREATER_THAN
-            Threshold: 80
-            ThresholdType: PERCENTAGE
-          Subscribers:
-            - SubscriptionType: SNS
-              Address: !Ref NotificationTopic
-        - Notification:
             NotificationType: FORECASTED
             ComparisonOperator: GREATER_THAN
             Threshold: 100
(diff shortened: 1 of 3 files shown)


thomasyu888 (approved):
Lgtm

templates/budget.yaml
```diff
@@ -150,7 +150,7 @@ Resources:
         - Notification:
             NotificationType: ACTUAL
             ComparisonOperator: GREATER_THAN
-            Threshold: 100
+            Threshold: 200
```
zaro0508, line 153: hmm, seems odd.  is this saying you want to be notified only if the cost is 200% over the budget?

BrunoGrandePhD: I want to be notified when the forecasted cost passes 100%, when the actual cost passes 100%, and when the actual cost passes 200%. 

The reason for the new 200% threshold is to account for normal spikes in cost, which might happen a few times per year. When I get a notification and confirm that the spike is normal, I want to still be notified if the cost continues to grow much beyond the 100% threshold. I don't want to permanently increase the threshold because most months should be below it.

zaro0508 (approved):

BrunoGrandePhD:
I deleted the budgets from all three AWS accounts. I can now merge this PR.

PR #119, opened 2022-06-22
[WORKFLOWS-410] Disable Tower Fusion in light of persistent issues

BrunoGrandePhD:
Unfortunately, Fusion hasn't stopped orphaned EBS volumes from appearing in the account. It's also been causing at least two unprecedented issues, which has led me to decide to disable it.


thomasyu888 (approved):

thomasyu888:
Unfortunate...

wpoehlm (approved):

PR #120, opened 2022-07-19
Added configuration files to onboard new users

wpoehlm:
To onboard new users, I've added configuration files for Sophia Jobe, Verena Chung, Jared Hendrickson, and Wei-An Chen.


wpoehlm:
@BrunoGrandePhD please note that IT has not yet added them to the Jumpcloud groups, so if this is necessary before merging please hold off until I comment again.

thomasyu888 (approved):
LGTM - will let Bruno merge when he's back

config/projects-prod/verena-chung-project.yaml
```diff
@@ -0,0 +1,56 @@
+# (REQUIRED) Step 0: Duplicate this file and rename using the updated stack name (see Step 1)
+
+# (REQUIRED) Step 1: Update the stack name below by replacing "example" with your project name
+#                    Note that the stack name cannot contain more than 32 characters
+stack_name: verena-chung-project
+
+stack_tags:
+  # (REQUIRED) Step 2: Update all four values below
+  Department: CNB
+  Project: challenge
```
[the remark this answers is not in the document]
vpchung: @wpoehlm yes, those are correct - thanks!

wpoehlm:
@BrunoGrandePhD users have been added to the JumpCloud groups, so this should be ready to merge pending your approval

config/projects-prod/sophia-jobe-project.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+parameters:
+
+  S3ReadWriteAccessArns:
+    # (REQUIRED) Step 3: Replace the email below with your '@sagebase.org' address
+    - arn:aws:sts::563295687221:assumed-role/AWSReservedSSO_Developer_baa6fed639faf5e7/sophia.jobe@sagebase.org
+
+    # (Optional) Step 4: Uncomment and update the following line(s) to grant additional users with read/write access
+    # - arn:aws:sts::563295687221:assumed-role/AWSReservedSSO_Developer_baa6fed639faf5e7/tess.thyer@sagebase.org
+    # - arn:aws:sts::563295687221:assumed-role/AWSReservedSSO_Developer_baa6fed639faf5e7/william.poehlman@sagebase.org
```
BrunoGrandePhD, line 22: Do you want to add yourself to these `projects-prod` stacks so you can help debug issues during the workshop?

BrunoGrandePhD (changes requested):
I'm appr

config/projects-prod/sophia-jobe-project.yaml
wpoehlm: I like this idea theoretically, but if these projects are going to be potentially used for real data in the future then it might not be appropriate for me to have direct access depending on the project?

BrunoGrandePhD: I don't think we should have Tower projects named after individuals. We don't have enough AWS Batch slots for compute environments to support this approach. Once the workshop is done, we should decommission these individual-based projects in favor of grant-based projects. We don't have enough time to identify the latter, so we can stick with the former just for tomorrow.

BrunoGrandePhD: You can specify that no PHI data should be used so you can provide support during the workshop.

wpoehlm: In the case of Jared and Wei-An, I'd like these to be long term projects they can use. Would you recommend that I remove the individual accounts and add them to mine, or can we keep those as is (and still use them after the workshop)?

BrunoGrandePhD: It depends on how you want to organize the workflow runs. I'm less concerned about running into the AWS Batch limit on the STRIDES account since it has very few Tower projects so far. If there will be overlap in the data used, then a single Tower project might make more sense so you don't have to duplicate data across S3 buckets.

wpoehlm:
I've given myself read-write access to Verena and Sophia's project for the workshop. I would prefer to keep Jared and Wei-An's projects separate, and I have read/write access on them as well, so if you're not worried about batch limits on strides then we can leave it?

BrunoGrandePhD (approved):

PR #122, opened 2022-08-02
Offboard Haley from GENIE

thomasyu888:


BrunoGrandePhD (approved):

BrunoGrandePhD:
Unfortunately, the merge didn't work. As far as I can tell from the CloudFormation event log, the process of updating the KMS key that's created for each Tower project requires permission to the `kms:UntagResource` action, which isn't listed [here](https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/KMS/kms-key.yaml#L42-L53). I've opened a [pull request](https://github.com/Sage-Bionetworks/aws-infra/pull/350) on aws-infra, but this won't get reviewed until next week. I'm going to revert this change and open a new PR that depends on my PR so we don't lost track of this.

thomasyu888:
Thanks Bruno!

PR #125, opened 2022-08-12
[WORKFLOWS-394] Add `TowerViewer` roles ahead of sandbox deprecation

BrunoGrandePhD:
WIP


thomasyu888 (approved):
Lgtm. Just re-confirming that all the amp-ad resources should fall under the strides roles.

BrunoGrandePhD:
I'm relying on the resource-based policies on the buckets to grant S3 access. So, these roles need to be in the same account as the bucket. Otherwise, you have need to set up cross-account access, which requires the IAM roles to gain S3 permissions, which gets complicated fast.

PR #126, opened 2022-08-15
Create nf-ntap-biobank-jineta project

jaybee84:
I am creating this project to process additional files from the JHU-biobank. I tried launching runs from the existing ntap-add5-project but have been running into access related issues since the Synapse project has an upgraded access restriction.
Since more files are expected in this project later in the year as well as over the next few years, I think this project will be useful in running those in the future as well.

### DIFF
--- a/config/projects-prod/nf-ntap5-biobank-jineta.yaml
+++ b/config/projects-prod/nf-ntap5-biobank-jineta.yaml
@@ -20,7 +20,7 @@ parameters:
     
   # (Optional) Step 5: Uncomment and update the following line(s) to grant additional users with read-only access
   # S3ReadOnlyAccessArns:
-     - arn:aws:sts::563295687221:assumed-role/AWSReservedSSO_Developer_baa6fed639faf5e7/sasha.scott@sagebase.org
+    - arn:aws:sts::563295687221:assumed-role/AWSReservedSSO_Developer_baa6fed639faf5e7/sasha.scott@sagebase.org
   #   - arn:aws:sts::563295687221:assumed-role/AWSReservedSSO_Developer_baa6fed639faf5e7/jason.hwee@sagebase.org
 
   # (Optional) Step 6: Uncomment the following line to disable the feature allowing Synapse to index files
--- a/config/projects-prod/nf-ntap5-biobank-jineta.yaml
+++ b/config/projects-prod/nf-ntap5-biobank-jineta.yaml
@@ -1,0 +1,53 @@
+stack_name: nf-ntap5-biobank-jineta
+
+stack_tags:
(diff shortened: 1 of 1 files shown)


thomasyu888:
To clarify, is this because you can only have one secret per NF project so you need to be logged in as a different Synapse user?

jaybee84:
This is mainly due to the following reasons:
1. I am able to log into ntap-add5-project but when I try to run synstage I am getting an access restriction error possibly due to the new ARs in this project. So I was hoping to have a separate instance to troubleshoot this. Since Bruno had set up the original project and is out today, it is hard to pinpoint where things are getting stuck. 
2. This project has some additional files that I would like to process without adding to the DPE workload/roadmap for this year. Since this processing is time sensitive, it would be great for me to just go ahead an process them separately :)

tthyer:
@jaybee84 I've looked at this and it looks ok (FWIW since I'm pretty out of date with nextflow-infra at this point). One question I have: the cost center "NTAP-JHU / 30110" doesn't appear in the list of valid codes stored in aws-infra. Have we departed from the practice of having codes be one from the valid list, or are you following a separate process to add the code to that list?

config/projects-prod/nf-ntap5-biobank-jineta.yaml
```diff
@@ -0,0 +1,53 @@
+stack_name: nf-ntap5-biobank-jineta
+
+stack_tags:
+  # (REQUIRED) Step 2: Update all four values below
+  Department: SCCE
+  Project: Neurofibromatosis
+  OwnerEmail: jineta.banerjee@sagebase.org
+  CostCenter: NTAP-JHU / 30110 # Valid values here: https://github.com/Sage-Bionetworks/aws-infra/tree/master/templates/tags
```
jaybee84, line 8:
```suggestion
  CostCenter: NTAP NF Addendum 5 / 301101 # Valid values here: https://github.com/Sage-Bionetworks/aws-infra/tree/master/templates/tags
```

jaybee84: Thanks for catching that Tess. In my previous version, I could not find the appropriate code in the Program Codes json, but looked up the value in one of Bruno's older PRs. It looks like that was updated at a later date. 

Also, I see now that it is present in the `ProgramCodesOther.json` but not in `ProgramCodes.json`

jaybee84:
Update: I think the error due to access restriction for the account authorized for the ntap-add5-project has been resolved, thanks to Jay and the governance team :)

jaybee84:
just checking to see if anything else is required for this PR to be merged in.. let me know..

thomasyu888 (approved):

PR #128, opened 2022-08-16
Rename robert-allaway-project to robert-allaway-project.yaml

allaway:
@BrunoGrandePhD - Not sure how I got this wrong this last time, but I noticed that my project workspace wasn't showing up in Tower. I am guessing this is why... 🤣


BrunoGrandePhD:
Your text editor really doesn't like file extensions. This [happened](https://github.com/allaway/nextflow-infra/commit/0e2e8acf50b1fe71a2dcb38be9f49b0c7def64b3) in February as well. 😆

allaway:
😄 I think it's the github web text editor that has done this to me both times. I don't use it much but when I do... 😬

PR #130, opened 2022-08-24
Shell script for configuring SSH via SSM

BrunoGrandePhD:
I often find myself having to remotely access EC2 instances using SSM. I typically want to use SSH because I either want to copy files off of the EC2 instance using SCP or edit files using VS Code (with the SSH Remote extension). This requires me running through the same steps every time. In light of this, I decided to capture those commands in a shell script that can automate the setup and minimize mistakes.

Here's an example command and associated output:

```console
❯ bash setup-ssh-via-ssm.sh i-0421aea79957b510c

Running the following shell script on i-0421aea79957b510c...
  cd
  mkdir -p .ssh;
  echo [...] > .ssh/authorized_keys;
  sudo cp -r .ssh /home/ec2-user/;
  sudo chmod 700 /home/ec2-user/.ssh;
  sudo chmod 600 /home/ec2-user/.ssh/authorized_keys;
  sudo chown -R ec2-user:ec2-user /home/ec2-user/.ssh;

Starting session with SessionId: bruno.grande@sagebase.org-0d714049d8619edac


Exiting session with sessionId: bruno.grande@sagebase.org-0d714049d8619edac.


You should now be able to connect using 'ssh i-0421aea79957b510c'.
You can just use the SSH command for this instance going forward.
Connecting now...

Last login: Wed Aug 24 19:55:28 2022 from localhost

   __|  __|  __|
   _|  (   \__ \   Amazon Linux 2 (ECS Optimized)
 ____|\___|____/

For documentation, visit http://aws.amazon.com/documentation/ecs
17 package(s) needed for security, out of 43 available
Run "sudo yum update" to apply all updates.
[ec2-user@ip-172-22-0-82 ~]$ 
```

### DIFF
--- a/bin/setup-ssh-via-ssm.sh
+++ b/bin/setup-ssh-via-ssm.sh
@@ -2,12 +2,22 @@
 
 #
 # This script uploads your public SSH key ('id_ed25519' by default)
-# to a home directory ('ec2-user' by default) on an EC2 instance.
+# to the home directories of 'ec2-user' and 'ssm-user'. While both
+# users are configured, it's recommended to use 'ec2-user' for SSH.
 # Written by Bruno Grande.
 #
-# Usage:    bash setup-ssh-via-ssm.sh INSTANCE_ID
+# Script usage:
 #
-# Example:  bash setup-ssh-via-ssm.sh i-0d3ff331a7b7a7e8b
+#   bash setup-ssh-via-ssm.sh INSTANCE_ID [PUBLIC_SSH_KEY]
+#
+# Example commands:
+#
+#   # Generic usage:
+#   bash setup-ssh-via-ssm.sh i-0d3ff331a7b7a7e8b
+#
+#   # If you have another SSH key that you want to use,
+#   # which should match your SSH config file (see below):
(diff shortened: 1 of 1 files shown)


bin/setup-ssh-via-ssm.sh
```diff
@@ -0,0 +1,76 @@
...
+#
+#     Host i-* mi-*
+#         ProxyCommand sh -c "aws ssm start-session --target %h --document-name AWS-StartSSHSession --parameters 'portNumber=%p'"
+#         IdentityFile ~/.ssh/id_ed25519
+#         User ec2-user
+#
+
+# Command-line arguments and optional environment variables
+INSTANCE_ID="${1}"
+SSH_KEY_PUB="${SSH_KEY_PUB:-$HOME/.ssh/id_ed25519.pub}"
```
tthyer, line 46: nit: I have multiple SSH keys, so I might want to specify this without editing the file

BrunoGrandePhD: The public SSH key is now an optional positional argument!

PR #134, opened 2022-09-15
Add external HMS collaborators to `mc2-mcmicro-project`

BrunoGrandePhD:
- [ ] Add collaborators to `mc2-mcmicro-project` workspace manually

### DIFF
--- a/bin/configure-tower-projects.py
+++ b/bin/configure-tower-projects.py
@@ -215,7 +215,7 @@ class Projects:
                 email = match.group("session_name")
                 emails.append(email)
             else:
-                raise ValueError(
+                print(
                     f"Listed ARN ({arn}) doesn't follow expected format: "
                     "'arn:aws:sts::<account_id>:<role_name>:<email>'"
                 )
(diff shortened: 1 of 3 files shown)


bin/configure-tower-projects.py
```diff
@@ -215,7 +215,7 @@ def extract_emails(self, arns: Sequence[str]) -> List[str]:
                 email = match.group("session_name")
                 emails.append(email)
             else:
-                raise ValueError(
+                print(
```
thomasyu888, line 218: Will the new ARN format cause issues in this tower config script?

BrunoGrandePhD: The new ARN format will cause this print statement to output, but they will otherwise be ignored. That's why they'll have to be added to the Tower workspace manually, as per the checkbox in my PR description. This is intended to be a short-term fix until we can consolidate user information into a single config file as per WORKFLOWS-453.

thomasyu888: Ah yes - should have read the description. 🙃 Thanks Bruno!

templates/nextflow-tower-config-bucket.yaml
```diff
@@ -69,6 +69,8 @@ Resources:
               allow-list:
                 - "*@sagebase.org"
                 - "*@sagebionetworks.org"
+                - "artem.sokolov@gmail.com"
+                - "clarenceyapp111@gmail.com"
```
adamjtaylor, line 73: also add `jmuhlich@gmail.com`

PR #141, opened 2022-11-07
[WORKFLOWS-455] Harden separation in S3 access between Tower projects

BrunoGrandePhD:
At the time of writing, access between Tower projects is impossible due to KMS encryption. This PR builds on that by restricting access using identify and resource policies. There were two "paths" that I wanted to block:

1. One of the IAM roles that are automatically created by Tower Forge is attached to the `AmazonS3ReadOnlyAccess` managed policy, which would grant it access to all S3 buckets within the account. In theory, these roles aren't used to access project buckets, but it's better to be safe than sorry. To address this, I added a specific block against all roles created by Forge (_i.e._ prefixed with `TowerForge-`). 
2. The head and worker job roles created by the `tower-project` template were also attached to the overly permissive `AmazonS3ReadOnlyAccess` policy. I dropped this in favor of an inline policy (`TowerForgeExternalS3AccessPolicy`) that grants these roles with access to all S3 resources outside of the account. This should complement the access they are granted by the bucket policies.

### DIFF
--- a/templates/tower-project.yaml
+++ b/templates/tower-project.yaml
@@ -193,7 +193,7 @@ Resources:
         Statement:
           - Sid: GrantExternalS3Access
             Effect: Allow
-            Action: "*"
+            Action: "s3:*"
             Resource: "*"
             Condition:
               StringNotEquals:
--- a/templates/tower-project.yaml
+++ b/templates/tower-project.yaml
@@ -191,9 +191,13 @@ Resources:
       PolicyDocument:
         Version: '2012-10-17'
         Statement:
-          - Sid: GrantExternalS3Access
+          - Sid: ConditionalAmazonS3ReadOnlyAccess
             Effect: Allow
-            Action: "s3:*"
+            Action:
+              - "s3:Get*"
+              - "s3:List*"
+              - "s3-object-lambda:Get*"
+              - "s3-object-lambda:List*"
             Resource: "*"
             Condition:
               StringNotEquals:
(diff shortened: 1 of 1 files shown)


templates/tower-project.yaml
```diff
@@ -188,6 +184,25 @@ Resources:
...
+  TowerForgeExternalS3AccessPolicy:
+    Type: AWS::IAM::Policy
+    Properties:
+      PolicyName: TowerForgeExternalS3AccessPolicy
+      PolicyDocument:
+        Version: '2012-10-17'
+        Statement:
+          - Sid: GrantExternalS3Access
+            Effect: Allow
+            Action: "*"
```
thomasyu888, line 196: This seems to allow all AWS actions?

BrunoGrandePhD: Good catch!

thomasyu888 (approved):

BrunoGrandePhD:
Currently running an `nf-core/rnaseq` test before merging this.

templates/tower-project.yaml
```diff
@@ -188,6 +184,29 @@ Resources:
...
+    Type: AWS::IAM::Policy
+    Properties:
+      PolicyName: TowerForgeExternalS3AccessPolicy
+      PolicyDocument:
+        Version: '2012-10-17'
+        Statement:
+          - Sid: ConditionalAmazonS3ReadOnlyAccess
+            Effect: Allow
+            Action:
+              - "s3:Get*"
```
thomasyu888, line 197: Great addition - its too bad the `ManagedPolicyArns` doesn't take a mapping where you can specify the Condition.

BrunoGrandePhD:
The test job completed successfully, so I can merge this.

PR #142, opened 2022-11-08
[WORKFLOWS-461] Clean up Tower workspaces

BrunoGrandePhD:
In order to move forward on WORKFLOWS-461, I need to create some space for new compute environments in AWS Batch. This PR decommissions 4 Tower projects, which will free up 12 slots in AWS Batch. I also took the opportunity to remove the old sandbox roles (no longer available), delete extraneous comments, and remove mentions of Tess. 

Depends on #141

### DIFF
--- a/config/infra-dev/nextflow-ecs-task-definition.yaml
+++ b/config/infra-dev/nextflow-ecs-task-definition.yaml
@@ -28,7 +28,6 @@ parameters:
   TowerUserWorkspace: 'false'
   TowerRootUsers:
     - bruno.grande@sagebase.org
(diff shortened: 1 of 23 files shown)


wpoehlm (approved):
Since we are using strides for the foreseeable future, you can also remove config/projects-prod/amp-ad-project.yaml if it would help free up more space

BrunoGrandePhD:
Thanks for suggesting that, @wpoehlm! The `amp-ad-project` stack is already in read-only mode, so nothing else to do there.

thomasyu888 (approved):

PR #143, opened 2022-11-19
[WORKFLOWS-481] Parametrize TowerViewer assumed-role ARN prefix

BrunoGrandePhD:
A lot of files of being changed in this PR, but most of it was done with search and replace queries in VS Code (example below). It basically moves the common prefix for the TowerViewer assumed-role ARN to the Sceptre config. 

**Edit:** I'm not sure what's wrong with the pre-commit in the CI workflow, so I'll need to debug that too. 

- [ ] Debug pre-commit in CI workflow

```
Search:   arn:aws:sts::035458030717:assumed-role/AWSReservedSSO_TowerViewer_5919a1dd6875ede2/([\w@.]+)
Replace:  '{{stack_group_config.tower_viewer_arn_prefix}}/$1'
```

### DIFF
--- a/.github/workflows/aws-deploy.yaml
+++ b/.github/workflows/aws-deploy.yaml
@@ -12,6 +12,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
+        with:
+          python-version: '3.10'
       - uses: pre-commit/action@v2.0.2
 
   deploy:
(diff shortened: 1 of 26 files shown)


thomasyu888 (approved):
LGTM, I took a look at the pre-commit bug... It seems like it has to do with the 3.8 python default set in the pre-commit config vs the default 3.11.0 setup in GitHub actions but also the past actions didn't have an issue.  I wonder if pinning the python setup version would help.

BrunoGrandePhD:
@thomasyu888: Syncing the Python versions in both the pre-commit config and the GH Actions workflow did the trick!

PR #144, opened 2022-11-19
[IBCDPE-404] Remove unlisted participants in each workspace

BrunoGrandePhD:
This PR addresses an issue that got introduced when we reverted from managing access using teams back to using users (c77aac2da7ac253a187b6461b022d4307fc18e9e). While the teams functionality removed any unexpected team members, this feature had not been implemented yet for the users functionality. This PR addresses this gap. 

In practice, this should ensure that any users that are removed from Tower project stacks will also be removed from the corresponding Tower workspace. 

Depends on #143

### DIFF
--- a/bin/configure-tower-projects.py
+++ b/bin/configure-tower-projects.py
@@ -8,7 +8,7 @@ import os
 import re
 import time
 from collections import defaultdict
-from typing import Dict, Iterator, List, Optional, Sequence, Tuple
+from typing import Dict, Iterator, List, Optional, Sequence, Tuple, Set
 
 import boto3
 import yaml  # type: ignore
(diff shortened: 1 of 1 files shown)


bin/configure-tower-projects.py
```diff
@@ -410,11 +409,47 @@ def set_participant_role(self, part_id: int, role: str) -> None:
...
+        return participants
+
+    def list_owner_participant_ids(self) -> List[int]:
+        """List the participant IDs of any workspace owners
+
+        This will include the workspace creator.
+        """
+        participants = self.list_participants()
+        owners = [part for part in participants if part["wspRole"] == "owner"]
+        owner_ids = [owner["participantId"] for owner in owners]
```
thomasyu888, line 435: Nit: I can't think of why it would be advantageous to have duplicated participant ids.  It could be beneficial returning a set here.

```
set(owner["participantId"] for owner in owners)
```

bin/configure-tower-projects.py
```diff
@@ -205,8 +205,7 @@ def extract_emails(self, arns: Sequence[str]) -> List[str]:
             List[str]: List of email from the role session names
         """
         role_arn_regex = re.compile(
-            r"arn:aws:sts::(?P<account_id>[0-9]+):assumed-role/(?P<role_name>[^/]+)"
-            r"/(?P<session_name>[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,})"
+            r".*/(?P<session_name>[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,})"
```
BrunoGrandePhD, line 208: This change is needed in light of #143.

thomasyu888 (approved):
LGTM!  Just a minor comment - approving anyways.  Thanks!

bin/configure-tower-projects.py
BrunoGrandePhD: Done!

PR #148, opened 2022-11-24
[ORCA-105] Create `orca-dev-project` Tower project

BrunoGrandePhD:
In response to a discussion in https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/3, we might want to explore using the Tower-dev instance for testing purposes with Airflow.


thomasyu888 (approved):
Thanks! LGTM!

PR #151, opened 2022-12-09
Add ARN for htan-dev external account

adamjtaylor:
As discussed in [IBCSPRT-119](https://sagebionetworks.jira.com/servicedesk/customer/portal/5/IBCSPRT-119), I'm adding an ARN from the `htan-dev` account to enable cross-account transfer out of the `htan-project-tower-bucket` bucket.


adamjtaylor:
I've updated this to use a sts ARN that should allow the `htan-dev` profile in my `.aws/config` access, but can revert back to use the plain role ARN if needed

PR #155, opened 2023-02-02
Ensure that MC2 external users are added to Tower

BrunoGrandePhD:
I'm using existing functionality in the `configure-tower` script to ensure that the external users for the MC2 project are added to the Tower workspace. This is the quickest way of achieving that (other than manually adding them again). The better approach would be to add a new section in the YAML file (_e.g._ using `scepter_user_data`), but since we expect these stacks to be revamped soon to support custom AWS Batch compute environments, I opted against that for expediency.

This is a low-risk PR and one of our users is blocked, so I'm going to merge immediately, but I'm still open to feedback.


thomasyu888:
+1

PR #159, opened 2023-02-16
[IT-2366] Upgrade Tower to release 22.4.1

BrunoGrandePhD:
This PR upgrades Tower from release 22.1.5 to 22.4.1. One of the new Tower features is resource tags, so each compute environment is now tagged with the associated Tower project `stack_tags`. I also tweaked the compute environment to allow for additional head job resources. I made sure to update the request to Tower with the latest format (based on the Tower web client). I'm using more modern instance families (m5/r5/c5 instead of m4/r4/c4). For now, I'm leaving Wave and FusionV2 disabled, but we can explore these features later on. I enabled data persistence for the Redis container in the Tower ECS task definition to prevent token expiration issues between deployments (which cause the Tower service to restart). Finally, I took the opportunity to sync up the IAM policies with the latest ones in the Tower docs. 

I've deployed this branch to `nextflow-dev`, and I've performed some tests. Everything seems good except for the "resume workflows with different CEs" functionality, which doesn't quite work as expected ([source](https://sagebionetworks.jira.com/browse/IT-2366?focusedCommentId=165318)). Before merging this PR, I want to confirm that there aren't many (or any large) incomplete workflows that would be wasted due to cache invalidation.


bin/configure-tower-projects.py
```diff
@@ -240,6 +241,28 @@ def extract_users(self) -> Dict[str, Users]:
...
+        tags_per_project = dict()
+        for config in self.load_projects():
+            stack_name = config["stack_name"]
+            stack_tags = config.get("stack_tags", {})
+            stack_tags["TowerProject"] = stack_name
+            stack_tags.pop("OwnerEmail", None)
+            if "CostCenter" in stack_tags:
+                _, _, program_id = stack_tags["CostCenter"].rpartition("/")
+                stack_tags["CostCenter"] = program_id.strip()
+            print(stack_tags)
```
BrunoGrandePhD, line 259:
```suggestion
```

thomasyu888 (approved):
🙏 awesome - thanks for pushing this through.  Unfortunate about the workflow jobs not being able to be resumed.. approving first

thomasyu888:
The tower wave exploration can come as part of the ITCR docker execution epic, and we have the fusion V2 benchmarking for later.

bin/configure-tower-projects.py
```diff
@@ -239,6 +240,27 @@ def extract_users(self) -> Dict[str, Users]:
...
+        for config in self.load_projects():
+            stack_name = config["stack_name"]
+            stack_tags = config.get("stack_tags", {})
+            stack_tags["TowerProject"] = stack_name
+            stack_tags.pop("OwnerEmail", None)
+            if "CostCenter" in stack_tags:
+                _, _, program_id = stack_tags["CostCenter"].rpartition("/")
+                stack_tags["CostCenter"] = program_id.strip()
+            pattern = re.compile(r"[A-z0-9_-]{1,39}")
+            assert all(pattern.fullmatch(key) for key in stack_tags.keys())
```
thomasyu888, line 259: If this regex fails, does that mean the tag needs to be updated upstream?

BrunoGrandePhD: I changed the logic to replace stretches of invalid characters with underscores instead of raising an error.

PR #162, opened 2023-02-23
[WORKFLOWS-515] Limit to cheaper `8xlarge` instance types

BrunoGrandePhD:
As part of #161, I increased the number of available instance types for the Tower compute environments. Unfortunately, AWS Batch's allocation strategy tends to result in bigger instance types being used (_e.g._ `r6in.24xlarge` with 96 vCPUs and 768 GB of memory, costing over $8/hr). While this instance might make sense when a bunch of jobs are submitted, it's problematic when a single job is left running on it, resulting in poor instance utilization. 

In this PR, I'm limiting the instance types to `8xlarge` (32 vCPUs) to ensure more incremental scaling. The most expensive instance type in the updated list costs $2/hr, so it's significantly less "painful" to have a single job running on it. While I was refining the list, I dropped instance type subtypes that are more expensive (_e.g._ NVMe-enabled, EBS-optimized, or network-optimized) until we benchmark them. For example, it's possible that NVMe-enabled instance types complete jobs faster, resulting in lower costs overall despite the higher cost per hour.

### DIFF
--- a/bin/configure-tower-projects.py
+++ b/bin/configure-tower-projects.py
@@ -15,7 +15,7 @@ import yaml  # type: ignore
 from sagetasks.nextflowtower.client import TowerClient
 
 # Increment this version when updating compute environments
-CE_VERSION = "v8"
+CE_VERSION = "v9"
 
 REGION = "us-east-1"
 ORG_NAME = "Sage Bionetworks"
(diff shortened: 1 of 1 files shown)


bin/configure-tower-projects.py
```diff
@@ -31,15 +31,26 @@
...
+    # 4xlarge instance (16 vCPUs)
+    "c6a.4xlarge", "c5a.4xlarge", "c6i.4xlarge", "m5a.4xlarge", "m6a.4xlarge",
+    "m6i.4xlarge", "r5a.4xlarge", "r6a.4xlarge", "r6i.4xlarge",
+    # 8xlarge instance (32 vCPUs)
+    "c6a.8xlarge", "c5a.8xlarge", "c6i.8xlarge", "m5a.8xlarge", "m6a.8xlarge",
+    "m6i.8xlarge", "r5a.8xlarge", "r6a.8xlarge", "r6i.8xlarge"
 )
-# All GPU-enabled (NVIDIA-only), x86-64 (Intel-only) instance types
-# Excluding 'p3dn' and 'p4*' due to costs
-GPU_EC2_INSTANCE_TYPES = ("g4dn", "g2", "g3s", "p2", "g3", "p3")
```
BrunoGrandePhD, line 42: I'm deleting this since it's not being used for now.

thomasyu888 (approved):
👍🏼 Makes sense.  Thanks for this cost saving change!

PR #166, opened 2023-03-16
[IBCSPRT-159] Add shorter data expiration lifecycle rules

BrunoGrandePhD:
Some projects stage large files that don't need to remain in the scratch bucket for long. This PR introduces shorter data expiration lifecycle rules as a short-term solution until Nextflow improves their file management and/or fix the `cleanup` functionality.

### DIFF
--- a/templates/tower-project.j2
+++ b/templates/tower-project.j2
@@ -434,7 +434,7 @@ Resources:
             Status: !Ref EnableScratchDataExpiration
             ExpirationInDays: 30
             Prefix: work/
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
🎉 I was thinking that we could add a 1 day, but I think 3 days is actually fine. This gives us a little breathing room.

BrunoGrandePhD:
I added a single-day option. It doesn't hurt to have it in case users have workflows that finish quickly (_i.e._ within a few hours). Thanks for the suggestion!

PR #167, opened 2023-03-21
[ORCA-171] Create container registry credentials for Wave service

BrunoGrandePhD:
In preparation of using Wave, we need to authenticate all requests to container registries. I identified the 4 most commonly used registries: Docker Hub, Quay.io, GHCR, and ECR. The first three have their credentials stored in LastPass, and I manually ported them over to Secrets Manager (documented in `CONTRIBUTING.md`). The ECR credentials are account-specific, so I authored a CloudFormation stack that creates an IAM user and grants it a minimal set of ECR permissions. I haven't tested whether this works with Wave yet. 

It hurts having stacks that are so similar in the various `infra-*` subfolders, but I don't see any way around it because each account has its own administrator role that needs access to the `workflows-infra` KMS key. 

**Edit:** I intentionally did not update the Tower configuration script to use these credentials since that should be tackled in a separate PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -117,12 +117,18 @@ The Nextflow Tower credentials (_i.e._ access tokens) were created manually usin
 
 ### AWS Secrets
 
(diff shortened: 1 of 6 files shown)


thomasyu888 (approved):
🛸 LGTM.  Going to approve, but I had a question:

Just so I'm understanding, does this mean that private container repositories can only be accessed if the repositories give our 4 service users access?

BrunoGrandePhD:
Good question! As a Tower user, the easiest way of allowing Tower to pull private container images is to grant one of these service accounts read-access to the relevant container repositories. This isn't a perfect solution, because as Brad found out, you can't add collaborators on a container repository on Docker Hub if you're not paying a subscription. 

Another method is for the user to add their own container registry credentials to the Tower workspace. This is complicated by the fact that users don't have edit-access to Tower workspace credentials. This is intentional because we decided early on that we want to control the Tower workspace from this repository. We could revisit this decision to improve the user experience for using private containers. 

None of this is documented because I don't want users to access private containers before we investigate how Wave handles them. Once we are comfortable with that, we can document what I've shared above.

PR #169, opened 2023-03-24
Update sceptre template_path

zaro0508:
The Sceptre template_path key has been deprecated[1], update with the `template` key.

[1] https://docs.sceptre-project.org/dev/docs/stack_config.html?highlight=template_path#template-path


thomasyu888 (approved):
💯 Thanks Khai for making this change!  I think `configure-tower-projects.py` is the only script impacted by this, but Ill let Bruno take another closer peak.

PR #170, opened 2023-03-24
[IT-2376] Switch from LaunchConfiguration to LaunchTemplate

zaro0508:
EC2 Launch configurations has been deprecated[1] in favor of LaunchTemplate.

[1] https://aws.amazon.com/blogs/compute/amazon-ec2-auto-scaling-will-no-longer-add-support-for-new-ec2-features-to-launch-configurations/

### DIFF
--- a/templates/nextflow-ecs-cluster.yaml
+++ b/templates/nextflow-ecs-cluster.yaml
@@ -190,7 +190,9 @@ Resources:
     Properties:
       VPCZoneIdentifier:
         - !Ref SubnetId
-      LaunchConfigurationName: !Ref EcsLaunchTemplate
+      LaunchTemplate:
+        LaunchTemplateId: !Ref EcsLaunchTemplate
+        Version: !GetAtt EcsLaunchTemplate.LatestVersionNumber
       MinSize: '0'
       MaxSize: !Ref AutoscalingGroupMaxSize
       DesiredCapacity: !Ref AutoscalingGroupDesiredCapacity
--- a/templates/nextflow-ecs-cluster.yaml
+++ b/templates/nextflow-ecs-cluster.yaml
@@ -117,38 +117,40 @@ Resources:
       RetentionInDays: 14
 
   EcsLaunchTemplate:
-    Type: AWS::AutoScaling::LaunchTemplate
+    Type: AWS::EC2::LaunchTemplate
     Properties:
-      ImageId: !Ref EcsAmiId
-      InstanceType: !Ref EcsInstanceType
-      IamInstanceProfile: !Ref EcsInstanceProfile
-      SecurityGroups:
-        - !Ref EcsSecurityGroupId
-      BlockDeviceMappings:
-        - DeviceName: !Ref RootDeviceName
-          Ebs:
-            VolumeSize: !Ref RootEbsVolumeSize
-            VolumeType: !Ref EbsVolumeType
-      UserData:
-        Fn::Base64: !Sub |
-          MIME-Version: 1.0
-          Content-Type: multipart/mixed; boundary="==BOUNDARY=="
+      LaunchTemplateData:
+        ImageId: !Ref EcsAmiId
+        InstanceType: !Ref EcsInstanceType
+        IamInstanceProfile:
+          Arn: !GetAtt EcsInstanceProfile.Arn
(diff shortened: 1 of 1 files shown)


templates/nextflow-ecs-cluster.yaml
```diff
@@ -116,6 +116,40 @@ Resources:
       LogGroupName: !Join ['-', [ECSLogGroup, !Ref 'AWS::StackName']]
       RetentionInDays: 14
 
+  EcsLaunchTemplate:
```
zaro0508, line 119: This `LaunchTemplate` configuration sets the exact same parameters as the existing `LaunchConfiguration` except for [AssociatePublicIpAddress](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-autoscaling-launchconfiguration.html#cfn-autoscaling-launchconfiguration-associatepublicipaddress) which is not supported in LaunchTemplate.  Also the launch template defaults to a non-public address setting when set to a non-default subnet which is what is what is being done in the `EcsAutoScalingGroup` resource.

from the docs..
```
If the instance is launched into a nondefault subnet,
the default is not to assign a public IPv4 address
```

thomasyu888: Thanks for the info. The changes make sense, but should we remove the LaunchConfiguration?

thomasyu888 (approved):
🚀 LGTM!

templates/nextflow-ecs-cluster.yaml
```diff
@@ -116,6 +116,40 @@ Resources:
...
+          repo_upgrade: security
+
+          write_files:
+          - path: /etc/ecs/ecs.config
+            content: |
+              ECS_CLUSTER=${EcsClusterName}
+              ECS_IMAGE_PULL_BEHAVIOR=prefer-cached
+            append: true
+
+          --==BOUNDARY==--
```
BrunoGrandePhD, line 151: When I tested this, I ran into a failure. I think you meant to use `AWS::EC2::LaunchTemplate` as the type as opposed to `AWS::AutoScaling::LaunchTemplate`. After making that change, I had to nest the existing properties under `LaunchTemplateData`. I also had to add the `Arn` key under `IamInstanceProfile` as well as switch from `SecurityGroups` to `SecurityGroupIds`. Here's the version that I was able to deploy. Could you sanity-check it? 

```suggestion
Type: AWS::EC2::LaunchTemplate
    Properties:
      LaunchTemplateData:
        ImageId: !Ref EcsAmiId
        InstanceType: !Ref EcsInstanceType
        IamInstanceProfile:
          Arn: !GetAtt EcsInstanceProfile.Arn
        SecurityGroupIds:
          - !Ref EcsSecurityGroupId
        BlockDeviceMappings:
          - DeviceName: !Ref RootDeviceName
            Ebs:
              VolumeSize: !Ref RootEbsVolumeSize
              VolumeType: !Ref EbsVolumeType
        UserData:
          Fn::Base64: !Sub |
            MIME-Version: 1.0
            Content-Type: multipart/mixed; boundary="==BOUNDARY=="

            --==BOUNDARY==
            Content-Type: text/cloud-config; charset="us-ascii"

            #cloud-config
            repo_update: true
            repo_upgrade: security

            write_files:
            - path: /etc/ecs/ecs.config
              content: |
                ECS_CLUSTER=${EcsClusterName}
                ECS_IMAGE_PULL_BEHAVIOR=prefer-cached
              append: true

            --==BOUNDARY==--
```

BrunoGrandePhD (changes requested):

zaro0508:
> 🚀 LGTM! @BrunoGrandePhD / @zaro0508 do you think that at some point, the CI/CD should push changes to the dev infra first so they can be tested on dev, then something else would trigger the prod deployment (e.g. A release tag)?

Yes, the matrix strategy currently in the CI setup is not ideal for creating dependent pipelines.  I'm planning to break up the matrix build to fix issue IT-2357 which will put us on a path to easily test deployments to dev before prod. 

The first step is PR https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/pull/172 which will catch a bunch of these initial issues from my PRs.

BrunoGrandePhD (approved):

PR #171, opened 2023-03-24
[IT-2376] Remove LaunchConfiguration

zaro0508:
We can remove LaunchConfiguration after switching to LaunchTemplate

depends on #170


thomasyu888 (approved):
💯 thanks!

BrunoGrandePhD:
Is there a reason why this is separate from #170?

BrunoGrandePhD (commented):
Is there a reason why this is separate from #170?

zaro0508:
> Is there a reason why this is separate from #170?

The difference is an in place replace vs an add then remove workflow.  I prefer the latter when changing resources resources that contain dependencies, it makes it easier to identify issues and rollback if needed.

BrunoGrandePhD (approved):

PR #173, opened 2023-03-27
[ORCA-143] Enable Wave for Nextflow Tower

BWMac:
This PR enables Wave in our Nextflow Tower Instances by adding the necessary environment variables. Coupled with #167, users should be able to create compute environments that use the Wave service to run workflows with docker containers including private containers to which their workspace credentials have access. In order to validate these changes, nf-core/rnaseq workflows were run:

- With Wave enabled and all containers public
- With Wave and Fusion V2 enabled and all containers public
- With Wave enabled and one container private

Discussions about how to enable Wave across the board or by default should follow merging this PR, and additional documentation/communication to users will be needed to help them navigate using Wave for their workflow runs.


BrunoGrandePhD (approved):
I'll approve, but we should note that this doesn't enable Wave for our users. It only makes it possible to create compute environments with Wave enabled. If we want to enable Wave across the board, we'll need to update the Tower configuration script, but I think we should have a discussion about that first.

BWMac:
@BrunoGrandePhD @thomasyu888 Since this doesn't _necessarily_ change the way that users interact with Tower, could we merge this PR and then have a discussion about enabling Wave across the board and the documentation that will be needed to help users navigate that larger change?

thomasyu888:
Thanks @BrunoGrandePhD - I totally spaced on that.

Let's merge this through so we can strategize how to make wave available to our users AND merge the other PRs that Khai has contributed. I think it is timely with our meeting with Khai this week to discuss compute environments.

PR #174, opened 2023-03-27
[WORKFLOWS-521] Check Redis append-only file first

BrunoGrandePhD:
Now that we are persisting the Redis cache between deployments, we are running into corruption issues when we re-deploy Tower. This draft PR demonstrates one way of solving the issue, but I would prefer preventing it altogether. 

In this PR, we are automatically running the `redis-check-aof` command on launch (like the migrate database command) before launching Redis. I'm not sure if the behavior is defined if you have an existing Redis instance accessing this file while it's being fixed by another process. This is largely why I would prefer preventing this issue. 

**Edit:** I test this, and it seems to work. Also, Seqera didn't respond with anything particularly useful, so I'm okay with moving forward with this stop-gap solution until we work on a more permanent approach.

### DIFF
--- a/templates/nextflow-ecs-task-definition.j2
+++ b/templates/nextflow-ecs-task-definition.j2
@@ -174,6 +174,8 @@ Resources:
           EFSVolumeConfiguration:
             FilesystemId: !Ref EfsFileSystemId
       ContainerDefinitions:
+        # The following container definition is a stop-gap solution for
+        # https://sagebionetworks.jira.com/browse/WORKFLOWS-521
         - Name: !Sub '${RedisContainerName}-CheckAOF'
           Image: !Ref RedisContainerImage
           Memory: 2000
--- a/templates/nextflow-ecs-task-definition.j2
+++ b/templates/nextflow-ecs-task-definition.j2
@@ -174,10 +174,30 @@ Resources:
           EFSVolumeConfiguration:
             FilesystemId: !Ref EfsFileSystemId
       ContainerDefinitions:
+        - Name: !Sub '${RedisContainerName}-CheckAOF'
+          Image: !Ref RedisContainerImage
+          Memory: 2000
+          Cpu: 0
+          EntryPoint:
+            - redis-check-aof
+          Command:
+            - --fix
+          MountPoints:
+            - ContainerPath: /data
+              SourceVolume: !Ref EfsVolumeName
+          LogConfiguration:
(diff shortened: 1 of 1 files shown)


thomasyu888:
@BrunoGrandePhD should we have a conversation with IT about this on Wednesday when we meet?

templates/nextflow-ecs-task-definition.j2
```diff
@@ -174,10 +174,34 @@ Resources:
           EFSVolumeConfiguration:
             FilesystemId: !Ref EfsFileSystemId
       ContainerDefinitions:
+        - Name: !Sub '${RedisContainerName}-CheckAOF'
```
zaro0508, line 177: maybe add a comment to indicate that this is a workaround for some jira issue?

templates/nextflow-ecs-task-definition.j2
```diff
@@ -174,10 +174,34 @@ Resources:
...
+          Memory: 2000
+          Cpu: 0
+          Essential: false
+          EntryPoint:
+            - /bin/sh
+          # The `yes` command is needed because `redis-check-aof` has an
+          # interactive prompt and doesn't have an option to disable it
+          Command:
+            - -c
+            - "yes | redis-check-aof --fix appendonly.aof"
```
zaro0508, line 188: i'm wondering if this is at all related? https://github.com/RediSearch/RediSearch/issues/394

thomasyu888 (approved):
🚀  Excellent!  Thanks for the quick fix!

BrunoGrandePhD:
I'm going to wait for @zaro0508 before merging.

zaro0508 (approved):
generally looks good, highly advise to add a comment.

templates/nextflow-ecs-task-definition.j2
BrunoGrandePhD: Good suggestion!

templates/nextflow-ecs-task-definition.j2
BrunoGrandePhD: It seems related. An alternative fix is to enable `aof-load-truncated yes` in the `redis.conf` file, but that would involve more work (_e.g._ create the config file, copy it to EFS, and mount it to the Redis container). I'll keep that approach as a backup if this PR fails to address the issue consistently.

PR #180, opened 2023-04-10
[IT-2361] Custom Batch compute environment

zaro0508:
Setup a custom batch compute environment to be used with nexflow
tower in a tower launch configuration so that we can avoid some of the
limitations of the tower forge configuration.

depends on https://github.com/Sage-Bionetworks/aws-infra/pull/392


templates/nextflow-launch.yaml
```diff
@@ -0,0 +1,204 @@
...
+    Type: CommaDelimitedList
+    Description: List of public Subnet Ids for the load balancer
+  MaxvCpus:
+    Type: Number
+    Description: The number of max CPUs for Batch compute environments
+    MinValue: 1
+    MaxValue: 16
+    Default: 1
+Resources:
+  UserManagedPolicy:
```
zaro0508, line 17: this is setup from the [configure AWS batch manually docs](https://install.tower.nf/22.3/advanced-topics/manual-aws-batch-setup/#create-a-user-policy) however I'm not sure how it's used.  I'm wondering whether this is needed at all?  would the **user** actually be using the [nextflow-launch-iam-policy.yaml](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/main/templates/nextflow-launch-iam-policy.yaml) instead?

templates/nextflow-launch.yaml
```diff
@@ -0,0 +1,204 @@
...
+        SpotIamFleetRole: !Ref SpotFleetRole
+        InstanceRole: !Ref InstanceRole
+        LaunchTemplate:
+          LaunchTemplateId: !ImportValue
+            'Fn::Sub': '${AWS::Region}-nextflow-ecs-cluster-EcsLaunchTemplate'
+          LaunchTemplateName: !ImportValue
+            'Fn::Sub': '${AWS::Region}-nextflow-ecs-cluster-EcsLaunchTemplate'
+          Version: !ImportValue
+            'Fn::Sub': '${AWS::Region}-nextflow-ecs-cluster-EcsLaunchTemplateLatestVersionNumber'
+  JobQueueOnDemand:
```
zaro0508, line 160: these on demand and spot queues are separate and requires users to specify which queue to put their jobs on.  I assume NF tower has a way for users to specify this either using the NF CLI or console?

thomasyu888:
Thanks @zaro0508 for this work.  Bruno is on vacation for the next 2 weeks, so I'd like to wait for him to be back to fully test this.

In the interim, I think that we need to have a system for triggering the production deployment rather than relying on the successful deployment to dev (like using github tags/releases).  The reason is this has a big impact on the way interact with Tower, so I'd want to go in and make sure everything works as expected by running test workflows on tower.

templates/nextflow-launch.yaml
thomasyu888: Correct, there is a way to do this within the console and it looks like this:

image

zaro0508:
By all means @thomasyu888, do what you think is best.  if you need IT help then please enter a jira issue and let us know.

templates/nextflow-launch.yaml
BrunoGrandePhD: I'm not sure if they meant "IAM user" or "Tower user" in this context. Either way, Tower users wouldn't be launching workflows directly on AWS Batch themselves. They would be launching them via Tower. I also don't see how this policy gets used later in the documentation, so I'm not sure we need it. 

Here's the general flow: 

1. A user submit a workflow using the Tower web UI or API
2. The Tower ECS service uses its IAM role to assume a project-specific role (the goal of #181), which has permissions to that project's Batch resources and S3 buckets. 
3. The project-specific role submits the Nextflow head job to an on-demand queue (we configure Tower so it knows which Batch queue to use).
4. The Nextflow head job uses its own role to submit the worker jobs to either the on-demand or spot queue.

templates/nextflow-launch.yaml
BrunoGrandePhD, line 1: We will need to deploy these resources once per Tower project, similar to what I described in my review for #181. I'll let you determine the best way to lay this out using Sceptre because I know `tower-project.j2` is already big enough as it is.

BrunoGrandePhD (changes requested):
I haven't taken a close look at the resources since this needs to be restructured to be created for each Tower project. Once we have that set up, we can test the deployment.

templates/nextflow-launch.yaml
zaro0508: > We will need to deploy these resources once per Tower project

i have moved the deployment of nextflow-launch.yaml into tower-project.j2

BWMac (approved):

PR #181, opened 2023-04-10

zaro0508:
Allow tower to use roles instead of a service user to access AWS resources.


templates/nextflow-tower-role.yaml
```diff
@@ -0,0 +1,150 @@
...
+  PrincipalArns:
+    Type: CommaDelimitedList
+    Description: Grant access to principals (accounts, groups, and users).
+    ConstraintDescription: List of ARNs (i.e. ["arn:aws:iam::011223344556:user/jsmith", "arn:aws:iam::544332211006:user/rjones"])
+  BucketArns:
+    Type: CommaDelimitedList
+    Description: Grant access to NF tower buckets.
+    ConstraintDescription: List of ARNs (i.e. ["arn:aws:s3:::BUCKET-01/*", "arn:aws:s3:::BUCKET-02/*"])
+Resources:
+  ManagedPolicy:
```
BrunoGrandePhD: These policies already exist here: [nextflow-forge-iam-policy.yaml](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/dev/config/common/nextflow-forge-iam-policy.yaml) and [nextflow-launch-iam-policy.yaml](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/dev/config/common/nextflow-launch-iam-policy.yaml). So, there's no need to deploy them again.

config/infra-dev/nextflow-tower-role.yaml
```diff
@@ -0,0 +1,11 @@
...
+  path: nextflow-tower-role.yaml
+stack_name: nextflow-tower-role
+stack_tags:
+  {{stack_group_config.default_stack_tags}}
+parameters:
+  PrincipalArns:
+    - "arn:aws:sts::035458030717:root"
+    - "arn:aws:sts::526515999252:assumed-role/AWSReservedSSO_Administrator_66bb518af6b2dfa2/bruno.grande@sagebase.org"
+  BucketArns:
+    - !stack_output infra-dev/nextlfow-tower-config-bucket.yaml::BucketArn
```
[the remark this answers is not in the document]
BrunoGrandePhD: This role would get created once per Tower project/workspace and would be granted permissions on the S3 buckets for each Tower project. This role does not need access to the Tower config bucket, which is only used by the Tower web app.

templates/nextflow-tower-role.yaml
```diff
@@ -0,0 +1,150 @@
...
+            Effect: Allow
+            Action:
+              - s3:*Object*
+            Resource: !Ref BucketArns
+  Role:
+    Type: "AWS::IAM::Role"
+    Properties:
+      ManagedPolicyArns:
+        - !Ref ManagedPolicy
+      AssumeRolePolicyDocument:
```
BrunoGrandePhD: Once this role is set up for each Tower workspace, I think it'll need to be "assumable" by the Tower ECS role. There are a few IAM roles related to ECS, so I'm not 100% sure which one needs to be trusted. Here's [my guess](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/8b8da9b40a9a82bbb2f18ad356ad684e07cd8216/templates/nextflow-ecs-service.yaml#L54-L79).

BrunoGrandePhD (changes requested):
As I mentioned in one of the comments, this IAM role would need to get deployed once per Tower project. It's intended to replace the [IAM user](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/8b8da9b40a9a82bbb2f18ad356ad684e07cd8216/templates/tower-project.j2#L94-L118) that's currently in the `tower-project.j2` template. However, I acknowledge that this template has grown to be quite large, so we can discuss how we should organize the resources if you think it's unmanageable. I know you're not a fan of nested stacks. I'll let you decide if you want to keep `tower-project.j2` as a "monolith" or not. 

If you need clarification, we can discuss on Slack.

BrunoGrandePhD (changes requested):
As I mentioned in one of the comments, this IAM role would need to get deployed once per Tower project. It's intended to replace the [IAM user](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/8b8da9b40a9a82bbb2f18ad356ad684e07cd8216/templates/tower-project.j2#L94-L118) that's currently in the `tower-project.j2` template. However, I acknowledge that this template has grown to be quite large, so we can discuss how we should organize the resources if you think it's unmanageable. I know you're not a fan of nested stacks. I'll let you decide if you want to keep `tower-project.j2` as a "monolith" or not. 

If you need clarification, we can discuss on Slack.

config/infra-dev/nextflow-tower-role.yaml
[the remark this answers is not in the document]
zaro0508: i have moved the resources into tower-project.j2

templates/nextflow-tower-role.yaml
zaro0508: changed to use existing policies.

templates/nextflow-tower-role.yaml
zaro0508: gave ECS role permission to assume tower role

thomasyu888 (approved):
LGTM!

PR #192, opened 2023-05-04
Deploy to production

BrunoGrandePhD:
Not sure what we should write in the description of these PRs. Maybe we should deploy to prod with versioned releases instead of merges into main. This way, we can generate changelogs automatically using GitHub. Just an idea.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,6 +23,7 @@ repos:
     rev: v1.2.0
     hooks:
       - id: mypy
+        args: [--implicit-optional]
   - repo: https://github.com/psf/black
     rev: 23.3.0
     hooks:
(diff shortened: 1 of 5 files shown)


thomasyu888 (approved):
✔️ LGTM!

View this thread/Jira ticket linked: https://sagebionetworks.slack.com/archives/CEFQD0KU1/p1681397761840869?thread_ts=1681397761.840869&cid=CEFQD0KU1. I had a similar thought as yours, but I didn't push harder on it. Perhaps we can revisit at a later date.

BrunoGrandePhD:
Yeah, let's try this method for now.

zaro0508:
As described in issue IT-2790, creating this PR is an optional step.  You also have the option to just directly push these merges to prod.

PR #198, opened 2023-06-07
[IBCDPE-587] adds bash script for syncing LENS reference files

BWMac:
This PR adds a bash script that can be used for reproducibly `sync`-ing the LENS Nextflow Workflow reference files to our `sage-igenomes` S3 bucket. The files and their structure within the subdirectories can be viewed [here](https://s3.console.aws.amazon.com/s3/buckets/sage-igenomes?region=us-east-1&prefix=LENS/&showversions=false). 

I followed the [example](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/dev/bin/mirror-igenomes.sh) provided by Bruno as much as I could, but there are a couple of differences.

1. There is no all-encompassing location where these reference files are stored. Instead, the LENS devs provide 3 bash scripts that are used to generate the reference files from different locations. These scripts are downloaded and executed in this script.
2. The file structure is different in that there are both files and folders present in the `prefix` directories used in this script, so I decided against `sync`-ing with longer prefixes as Bruno did in the example for simplicity.

I considered not including prefixes and the `sync` loop at all because I am not 100% certain that the prefixes won't change in the future.

After a follow-up discussion with iAtlas stakeholders, we elected to look closer at the reference file-downloading bash scripts used to set up the `reference` directory. I have made some updates to these scripts and included them in this PR. before running these scripts, it is assumed that you have followed the setup instructions provided by the LENS team [here](https://gitlab.com/landscape-of-effective-neoantigens-software/nextflow/modules/tools/lens/-/wikis/Running-LENS#raft-installation) and that the `raft` directory has been created.

Documentation of the other additional steps I needed in order to run these scripts is also a WIP.

NOTE: The `generate_mouse_references.sh` script is still a WIP, owing to the fact that the `starfusion` section needs to be updated. As a result, I have commented out the `download_mouse_references.sh` executing lines.


BWMac:
I made another [small commit](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/tree/d19d1ae3fd634b0275a5c7b03f686df4c72dd330) to this branch but the PR does not seem to be updating for some reason...

bin/mirror-lens-references.sh
```diff
@@ -0,0 +1,29 @@
...
+)
+
+echo "syncing LENS reference files"
+mkdir -p "./lens/references"
+wget -P ./lens https://gitlab.com/landscape-of-effective-neoantigens-software/nextflow/modules/tools/lens/-/wikis/uploads/b0d8a24036628bd9e6c860eb531919c3/download_general_references.sh \
+    https://gitlab.com/landscape-of-effective-neoantigens-software/nextflow/modules/tools/lens/-/wikis/uploads/2238c7f4af5c5499f298315bf5080ad7/download_human_references.sh \
+    https://gitlab.com/landscape-of-effective-neoantigens-software/nextflow/modules/tools/lens/-/wikis/uploads/bf210753e3724edf6db2543eb54c3de4/download_mouse_references.sh
+bash download_general_references.sh ./references
+bash download_human_references.sh ./references
+bash download_mouse_references.sh ./references
```
BrunoGrandePhD: Does this need to be `./lens/references/`?

bin/mirror-lens-references.sh
```diff
@@ -0,0 +1,29 @@
...
+
+for prefix in ${prefixes[*]}; do
+    aws s3 --region us-east-1 sync "./references/$prefix/" "s3://sage-igenomes/LENS/$prefix/" --acl public-read
```
BrunoGrandePhD: Same here. Maybe there just needs to be a `cd ./lens` somewhere before the `bash download_*` commands.

bin/mirror-lens-references.sh
BrunoGrandePhD, line 1: Can you calculate the total size of the files in S3 and include that number in a comment? This way, future users will know roughly how much disk space they need to run this script.

bin/mirror-lens-references.sh
BWMac: Thanks for catching this, testing this out also made me realize that `download_general_references.sh` expects an environment variable called `RAFT_PATH` which is basically the parent directory. 

I added a line `cd ./lens`, and then the bash scripts all expect the `references` directory to be present in that directory.

bin/mirror-lens-references.sh
BWMac: added in latest commit

bin/mirror-lens-references.sh
```diff
@@ -0,0 +1,34 @@
...
+    "viral/"
+)
+
+echo "Downloading LENS reference files"
+mkdir -p "./lens/references"
+cd ./lens
+wget -P . https://gitlab.com/landscape-of-effective-neoantigens-software/nextflow/modules/tools/lens/-/wikis/uploads/b0d8a24036628bd9e6c860eb531919c3/download_general_references.sh \
+    https://gitlab.com/landscape-of-effective-neoantigens-software/nextflow/modules/tools/lens/-/wikis/uploads/2238c7f4af5c5499f298315bf5080ad7/download_human_references.sh \
+    https://gitlab.com/landscape-of-effective-neoantigens-software/nextflow/modules/tools/lens/-/wikis/uploads/bf210753e3724edf6db2543eb54c3de4/download_mouse_references.sh
+export RAFT_PATH="."
```
thomasyu888: Also, did we want to include instructions for installing raft?  Is that necessary?

Might want to add a git clone?

BWMac: You don't need to install RAFT to download the reference files so I don't think that's necessary here, but the process outlined in the LENS docs leads to you mounting these refs in your RAFT instance, hence the reference to `RAFT_PATH` here.

thomasyu888: Thanks, I'm loooking in the shell scripts, and it does seem like there are dependencies you needed to install like:

* singularity
* Python + certain python packages

Did you experience that?

BWMac: Yes, the instructions [here](https://gitlab.com/landscape-of-effective-neoantigens-software/nextflow/modules/tools/lens/-/wikis/Running-LENS#populate-raft-global-references-directory) include those installations, I will add this link to the script as set-up instructions

BrunoGrandePhD (approved):

bin/lens/download_general_references.sh
```diff
@@ -0,0 +1,28 @@
...
+# BLASTP binary (for antigen.garnish)
+mkdir -p bin; cd bin
+wget https://ftp.ncbi.nlm.nih.gov/blast/executables/blast+/2.13.0/ncbi-blast-2.13.0+-x64-linux.tar.gz
+tar xvf *gz
+mv ncbi*/bin/blastp $REFERENCES_PATH/antigen.garnish
+cd $REFERENCES_PATH
+rm -rf bin
+
+# Make dummy_file - I have no idea why this was here in the first place
+touch dummy_file
```
thomasyu888, line 28: Should we remove this if we have no idea why it is here?

bin/lens/download_human_references.sh
```diff
@@ -0,0 +1,132 @@
...
+wget https://storage.googleapis.com/genomics-public-data/resources/broad/hg38/v0/Homo_sapiens_assembly38.fasta
+docker pull staphb/samtools:1.13
+docker run -v $PWD:/data staphb/samtools:1.13 samtools faidx Homo_sapiens_assembly38.fasta
+# EBV removal strategy from https://bioinformatics.stackexchange.com/a/14421
+keep_ids=($(awk '{print $1}' Homo_sapiens_assembly38.fasta.fai | grep -v chrEBV))
+docker run -v $PWD:/data staphb/samtools:1.13 samtools faidx -o Homo_sapiens.assembly38.no_ebv.fa Homo_sapiens_assembly38.fasta "${keep_ids[@]}"
+rm -f *.fasta*
+cd $HUMAN_REFERENCES_PATH
+
+# stopped here
```
thomasyu888: Stopped here?

bin/lens/download_human_references.sh
```diff
@@ -0,0 +1,132 @@
...
+
+
+# snpEff reference
+mkdir -p snpeff; cd snpeff
+docker pull resolwebio/snpeff:latest
+mkdir -p GRCh38.GENCODEv37; cd GRCh38.GENCODEv37
+wget https://gitlab.com/landscape-of-effective-neoantigens-software/nextflow/modules/tools/lens/-/wikis/uploads/430f9d80c841721499fbcec937b0f721/snpEff.config
+cp $HUMAN_REFERENCES_PATH/annot/gencode.v37.annotation.gtf.gz genes.gtf.gz
+gunzip genes.gtf.gz
+sudo ln $HUMAN_REFERENCES_PATH/fasta/Homo_sapiens.assembly38.no_ebv.fa sequences.fa
```
thomasyu888, line 72: Do you need sudo to link?

thomasyu888 (commented):
I know this is a draft, but some prelim comments.

bin/lens/download_human_references.sh
BWMac: forgot to remove!

bin/lens/download_human_references.sh
BWMac: not sure why, but in this case I did

bin/lens/download_general_references.sh
BWMac: it's in the list of files that Steven provided, I guess I would rather leave it in just in case it matters

thomasyu888 (approved):
🛠️ this looks great! Thanks!

PR #199, opened 2023-06-08
Fix MalformedPolicyDocument error in template

zaro0508:
It looks like most of the examples from the AWS IAM documented[1] shows that AWS princial parameter takes a list so we try that to fix the MalformedPolicyDocument that we are seeing in the deployment.

[1] https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html

### DIFF
--- a/mc2-mcmicro-dev-project.yaml
+++ b/mc2-mcmicro-dev-project.yaml
@@ -1,3 +1,4 @@
+---
 AWSTemplateFormatVersion: "2010-09-09"
 Description: >-
   Stack for provisioning the AWS resources needed for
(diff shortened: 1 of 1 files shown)


thomasyu888:
@zaro0508 is there a reason why you've added: [mc2-mcmicro-dev-project.yaml](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/pull/199/files#diff-2e7aa5ee911ec93accd1a50fd4f9a49ee2baff43e2391340563820055fdaffd8)

zaro0508:
this has been replaced with https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/pull/200

PR #205, opened 2023-06-08

zaro0508:
Allow tower to use roles instead of a service user to access AWS resources. This a 2nd attempt at PR #181, that one failed to deploy.

depends on #204


thomasyu888 (approved):
Nice!

PR #208, opened 2023-06-19
Increase intance type size

zaro0508:
The instance/ontainer that's running `redis-check-aof --fix appendonly.aof` is super duper slow which is causing the nexflow-tower service really slow to start.  We increase the instance size to see if it will run faster to unblock the redis/frontend/backend services from starting.


thomasyu888 (approved):
🔥 Nice!

PR #209, opened 2023-06-20
bump memory and add debug for redis check aof

zaro0508:
the redis-check-aof command is really so.  try giving it more memory to make it run faster.  Also add some debug statements so that we can log how long it takes to run.

### DIFF
--- a/templates/nextflow-ecs-task-definition.j2
+++ b/templates/nextflow-ecs-task-definition.j2
@@ -178,7 +178,7 @@ Resources:
         # https://sagebionetworks.jira.com/browse/WORKFLOWS-521
         - Name: !Sub '${RedisContainerName}-CheckAOF'
           Image: !Ref RedisContainerImage
-          Memory: 2000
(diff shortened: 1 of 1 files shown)


templates/nextflow-ecs-task-definition.j2
```diff
@@ -187,7 +187,7 @@ Resources:
           # interactive prompt and doesn't have an option to disable it
           Command:
             - -c
-            - "yes | redis-check-aof --fix appendonly.aof"
+            - echo 'start redis-check-aof' && 'yes | redis-check-aof --fix appendonly.aof' && echo 'redis-check-aof complete'
```
zaro0508, line 190: hope this is correct, got it from https://stackoverflow.com/a/61746646

thomasyu888: We'll deploy and see.  Looks ok

thomasyu888 (approved):
Thanks!

PR #210, opened 2023-06-20
Attempt to fix 'redis-check-aof' command

ConsoleCatzirl:
The current command is failing with `/bin/sh` failing to find an executable called `yes | redis-check-aof --fix appendonly.aof`.

Replace the single quotes with parantheses, which will fork a sub-process for the piped commands to run in, and wrap the entire YAML string in double quotes to prevent YAML from parsing the pipe operator.


thomasyu888 (approved):
Thanks!

PR #211, opened 2023-06-20
bump memory for redis check aof

zaro0508:
the redis-check-aof command is really so.  try giving it more memory to make it run faster.


thomasyu888 (approved):
Thanks!

PR #212, opened 2023-06-21
log timestap for redis-check-aof command

zaro0508:
Add timestamp to help us figure out how long it takes for redis-check-aof scan to execute.


thomasyu888 (approved):
Lgtm

PR #213, opened 2023-06-21
set assume role timeouts

zaro0508:
The redis-check-aof step on deployment takes significantly more time in the prod account than in the dev account.  We need to assume the aws role for longer period of time when deploying to prod.


thomasyu888 (approved):
Nice!

PR #214, opened 2023-06-21
EFS performance boost

zaro0508:
The EFS performance is saturated by the redis-check-aof command on the nexflow prod account.  The result is that it is taking 2hrs to execute the redis check. The EFS volume contains a few files one of them being the appendonly.aof file that's very large (~15GB).  We attempt to boost the EFS performance by configuring higher throughput for the EFS volume.  Hopefully this will allow redis-check-aof to scan the appendonly.aof file faster.

we could also set PerformanceMode=maxIO however the doc says 
`The performance mode can't be changed after the file system has been created.`

### DIFF
--- a/templates/nextflow-efs-file-system.yaml
+++ b/templates/nextflow-efs-file-system.yaml
@@ -28,6 +28,8 @@ Resources:
       BackupPolicy:
         Status: DISABLED
       Encrypted: true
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
👍🏼

PR #218, opened 2023-06-28
Allow GH OIDC access to KMS key

zaro0508:
We are now using github action OIDC for deployments and that role 
needs access to the workflow-infra KMS key to access secrets manager
resources.

### DIFF
--- a/config/infra-ampad/workflows-kms-key.yaml
+++ b/config/infra-ampad/workflows-kms-key.yaml
@@ -7,6 +7,7 @@ parameters:
   AccountAdminArns:
     - {{stack_group_config.sso_admin_role.arn}}
     - !stack_output_external workflows-nextflow-ci-service-account::ServiceRoleArn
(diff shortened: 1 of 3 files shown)


thomasyu888 (approved):
Thanks! LGTM

config/infra-ampad/workflows-kms-key.yaml
```diff
@@ -7,6 +7,7 @@ parameters:
   AccountAdminArns:
     - {{stack_group_config.sso_admin_role.arn}}
     - !stack_output_external workflows-nextflow-ci-service-account::ServiceRoleArn
+    - !stack_output_external workflows-infra-kms-key::KeyArn
```
ConsoleCatzirl, line 10: This should be the role arn assumed by the github action, right?

ConsoleCatzirl (changes requested):

PR #222, opened 2023-06-28
Remove GH service role

zaro0508:
We have switch to using a github OIDC role thus we no longer need the service role.


thomasyu888 (approved):
Nice!  Does this only impact the deployment or are there other potential impacts with this?

zaro0508:
i'm not aware of any other impacts, ohh we'll probably want to remove aws secrets from github after merging.

ConsoleCatzirl (approved):

PR #223, opened 2023-06-28
fix stack dependencies

zaro0508:
We need to deploy kms key updates before deploying resources that use those keys
otherwise the CI may encounter errors accessing secrets from the AWS secrets manager.


thomasyu888 (approved):
Thanks!  We will then push this to prod after

PR #224, opened 2023-06-29
[IBCDPE-586] Grant Rixing access to example project

rxu17:
**Purpose:** This grants `rixing.xu@sagebase.org` access to the example project on tower


thomasyu888 (approved):
➕ added!

PR #226, opened 2023-06-29
[IT-2968] Revert IT-2360 due to ampad deployment failure

zaro0508:
This reverts the change that caused ampad project deployments to fail.
We will need to figure out a different solution and try again.

PR #206, opened 2023-06-15
[IT-2360] Add in tower configuration to allow instance credentials

thomasyu888:
Leverage the feature implemented here #205

PR #229, opened 2023-07-04

zaro0508:
This is a 2nd attempt at PR #205 which was reverted in PR #226

We need to make the shared role into a common resource so that it
gets deployed to all accounts which will make the resource available
to tower project templates.

### DIFF
--- a/config/common/nextflow-ecs-service-shared.yaml
+++ b/config/common/nextflow-ecs-service-shared.yaml
@@ -1,0 +1,6 @@
+template:
+  path: nextflow-ecs-service-shared.yaml
+stack_name: nextflow-ecs-service-shared
+
+stack_tags:
+  {{stack_group_config.default_stack_tags}}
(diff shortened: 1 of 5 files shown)


thomasyu888:
@zaro0508 Do you know what the impact is to have ECS service role on the strides account?  Does this mean there is a separate nextflow tower instance running in the strides account?

zaro0508:
@thomasyu888 I don't really understand your question.  The only change that this PR will introduce is to deploy the ECS service role to all accounts (dev, prod, strides, etc..).  It does not add or remove any EC2 instances to any of those accounts.

thomasyu888:
Thanks @zaro0508 - so this role will be deployed across the different accounts, but will have access to the respective ECS clusters?

ie.
* dev ECS role -> dev ECS cluster
* prod ECS role -> prod ECS cluster
* strides ECS role -> prod ECS cluster

Do you think it would be worth having a "dev strides" sort of thing to test the "strides" like deployment but in the develop setting?

zaro0508:
no, not worth it

thomasyu888:
Thanks @zaro0508 , should we add this in: https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/pull/206?

zaro0508:
> Thanks @zaro0508 , should we add this in: #206?

done

thomasyu888 (approved):
Thanks Khai!

thomasyu888:
Hold off on merging this for now

PR #233, opened 2023-07-18
Add Sasha to jhu-biobank-nf-project.yaml

jaybee84:
FYI @ajs3nj


thomasyu888 (changes requested):
@jaybee84 I have no idea the impact of this change.

What I worry about is if the workspace gets deleted and a new workspace gets created and you lose all the data.

I suggest not renaming the workspace for now since this is time sensitive.

jaybee84:
@thomasyu888 sorry for the confusion. This is an old defunct workspace that we are reviving. I changed the name to avoid the exact kind of confusion that this PR has caused I think. 

1) This one is not time sensitive (the other one name "ntap-add5-project" is the one that is time sensitive)
2) The runs on this one are old test runs and ok to be deleted
3) If the workspace gets deleted and new one gets created that will be ok. Currently, neither I nor Sasha have read-write access to this workspace.

thomasyu888 (approved):

PR #239, opened 2023-09-15
Create nf-test.yaml

allaway:
Trying to debug why we cannot use external buckets on nextflow in nf-add5-project. This is a copy of htan-project, slightly modified to have the right ARNs.

### DIFF
--- a/config/projects-prod/nf-test.yaml
+++ b/config/projects-prod/nf-test.yaml
@@ -1,0 +1,25 @@
+template:
+  path: tower-project.j2
+stack_name: nf-project
+dependencies:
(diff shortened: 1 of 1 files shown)


thomasyu888 (changes requested):
Can we actually try this in the dev space? So project-dev.

allaway:
@thomasyu888, I was thinking the same thing  at first , but I was also wondering if that might affect testing (i.e. isn't that a differnt AWS account entirely?)

thomasyu888 (approved):

thomasyu888 (commented):
I will merge, then I'll have to push to prod branch to trigger the production ci/cd.

PR #241, opened 2023-09-19
Update policy to allow read write from external buckets

thomasyu888:
Update policy to allow read write from external buckets


zaro0508 (approved):
LGTM.  i'm not sure why you didn't just do `s3:*` but this should open up some more use cases for you.

PR #253, opened 2023-10-23
[IT-3184] optional self hosted redis container

zaro0508:
We have setup Elasticache redis in the dev environment.  In order to verify that Elasticache has taken over for the self hosted redis docker instance we need to disable it for the dev environment while still keeping it enabled for the prod environment.  This change allows the tower admin to enable or disable the self hosted redis docker resources per environment.


xschildw (approved):
LGTM

thomasyu888 (approved):

PR #266, opened 2023-11-06
[IT-3184] Elasticache for nextflow tower prod

zaro0508:
Switch nextflow tower from using a self hosted redis docker to AWS elasticache with redis service.  This has already been setup in the AWS nexflow-dev account, this is to enable it for the AWS nextflow-prod account.


thomasyu888 (approved):
Thanks. Do we need to drop the EFS scaling for prod too or is that already done in a separate PR?

zaro0508:
> Thanks. Do we need to drop the EFS scaling for prod too or is that already done in a separate PR?

one thing at a time is better.  we can scale back EFS after we verify elasticache is working in prod.

PR #269, opened 2023-11-14
[IT-3305] reduce EFS costs

zaro0508:
Change EFS throughput mode from provisioned to burstable to further reduce EFS cost.  Related to issue IT-3184


thomasyu888 (approved):
I will merge this into dev first and test.

PR #280, opened 2024-01-17
[IBCDPE-807] New PEGS Challenge workspace on tower-prod

jaymedina:
- [x] Create a new PEGS Challenge project workspace on tower-prod
- [x] Cost center is `12470001`
- [x] The following have been added with Read/Write access:
Rongrong Chai
Tom Yu
Jenny Medina
Brad MacDonald
Verena Chung
Gaia Andreoletti


config/projects-prod/pegs-challenge.yaml
```diff
@@ -0,0 +1,53 @@
+stack_name: pegs-challenge
+
+stack_tags:
+  Department: DNT
+  Project: PEGS Challenge
+  OwnerEmail: nextflow-admins@sagebase.org
+  CostCenter: 12470001 # Valid values here: https://github.com/Sage-Bionetworks/aws-infra/tree/master/templates/tags
```
jaymedina: In some of the other examples, I see tags formatted like `cost-center-name / 123456` - is the cost center name necessary or can I leave it like this?

thomasyu888: Good catch, this needs to be modified. I posted in #sageit

thomasyu888: @jaymedina : "PEGS Gene Environment / 124700"

config/projects-prod/pegs-challenge.yaml
thomasyu888, line 1: Can this follow the same file naming structure as the other projects?

`pegs-challenge-project`?

I think the code will actually fail if it doesn't follow that structure

thomasyu888 (changes requested):
Nice!  just left some comments

config/projects-prod/pegs-challenge.yaml
jaymedina: Thanks for posting - I've updated accordingly

config/projects-prod/pegs-challenge.yaml
jaymedina: OK, noted for the future

config/projects-prod/pegs-challenge-project.yaml
```diff
@@ -0,0 +1,53 @@
+stack_name: pegs-challenge
```
thomasyu888:
```suggestion
stack_name: pegs-challenge-project
```

thomasyu888 (approved):
✅  LGTM!

PR #272, opened 2023-12-08
[IBCDPE-764] Adds External User to `iatlas-project`

BWMac:
This PR adds an external user from ISB to the `iatlas-project` Tower workspace. I am requesting to merge this directly into `prod` so that the user can have access right away.

PR #285, opened 2024-02-05
[IT-3385] Bumps tower_version to Patched version

BWMac:
**Problem:**

When running workflows with a very large number of processes in Tower, workflows will fail.

**Solution:**

We have been in contact with Seqera about this issue for ~3 months now. They have finally published a patched version of the frontend container image we use to provide a fix. So, this PR bumps the `tower_version` we are using to this patched version.

I will post an announcement in the `tower-news` slack channel and give users a chance to see that this change is coming before merging.


config/config.yaml
```diff
@@ -1,20 +1,21 @@
 project_code: aws-workflows-nextflow-infra
-profile: {{ var.profile | default() }}
-region: {{ var.region | default("us-east-1") }}
+profile: {{var.profile | default()}}
```
zaro0508: any reason for all these formatting changes?

zaro0508 (approved):
thanks for the PR.

zaro0508 (changes requested):
ohh wait, this PR should be based on `dev` branch not `prod` branch

config/config.yaml
BWMac: My auto-formatter got me, reverted in the latest commit.

thomasyu888:
@zaro0508 since the dev branch is currently in the process of being updated to the newest version, we are doing a minor patch release to prod so we can move our work along.

thomasyu888 (approved):

zaro0508:
i think the better thing to do is patch dev -> patch prod -> move dev to latest

thomasyu888:
@BWMac Lets follow @zaro0508 's suggestion, hopefully that doesn't break dev to move backwards.

zaro0508:
this PR looks wrong

BWMac:
This PR was weird because my new branch was originally based on `prod` and then I changed the base to `dev`. There were changes in #267  and #272 which were pushed straight into `prod` and have now been merged into `dev` in this PR. `tower-dev` does appear to be working after the re-deploy, however.

PR #294, opened 2024-02-27
IT-3492: Remove resource name

xschildw:


thomasyu888 (approved):
thanks!  I'll approve, but I'll wait for Joni for their feedback

ConsoleCatzirl (commented):
I think this will resolve the deploy error, but I'm not sure if it will also disrupt any currently-running jobs.

xschildw:
If the update causes a resource replacement, I would expect it to cancel running jobs. Did we use the name to prevent that from happening?

thomasyu888:
> If the update causes a resource replacement, I would expect it to cancel running jobs. Did we use the name to prevent that from happening?

@xschildw , @jesusaurus not sure, but see this thread with @zaro0508 and I : https://sagebionetworks.slack.com/archives/CEFQD0KU1/p1707431159671579?thread_ts=1707179392.118219&cid=CEFQD0KU1.

ConsoleCatzirl (approved):

zaro0508 (changes requested):
DONT DO THIS!

PR #299, opened 2024-03-18
Update external users ARN

adamjtaylor:
Our collaborator Jacoson Jacobs at Emory has had to update his ARN due to a policy change on his end

As this is a change only in bucket policy I am proposing to make this change directly into production. 

@zaro0508 thank you in advance for your help getting this deployed.


adamjtaylor:
Withdrawing as also picks up other changes in dev

PR #300, opened 2024-03-18
Update external users ARN

adamjtaylor:
Our collaborator Jacoson Jacobs at Emory has had to update his ARN due to a policy change on his end

As this is a change only in bucket policy I am proposing to make this change directly into production.

@zaro0508 thank you in advance for your help getting this deployed.

Replaces #299


zaro0508:
> As this is a change only in bucket policy I am proposing to make this change directly into production.

No @adamjtaylor , please do not deviate from dev workflow.  please always target dev branch for PRs.

BWMac (approved):

zaro0508 (changes requested):
no

PR #304, opened 2024-04-02
[IT-3396] Switch to better jinja linter

zaro0508:
Switch out jinja linter.


BryanFauble (approved):
Awesome change! I really like to see the simplification of these GH pipelines

PR #307, opened 2024-04-29
Run Sceptre with prune option

zaro0508:
Sceptre should run with prune[1] option in CI workflow to allow it to remove stacks that are marked `obsolete`.

[1] https://docs.sceptre-project.org/latest/docs/cli.html#cmdoption-sceptre-launch-p


thomasyu888 (approved):
🔥 LGTM!

PR #311, opened 2024-06-10
[Workflows-549] Using IAM role in Tower to authenticate AWS services 

danlu1:
Problem: 

Currently, the IAM user is used by the Tower to access AWS. 

Solution:

Enable access to Tower workspace using an IAM Role

### DIFF
--- a/bin/configure-tower-projects.py
+++ b/bin/configure-tower-projects.py
@@ -536,7 +536,6 @@ class TowerWorkspace:
                 "provider": "aws",
                 "keys": {
                     "accessKey": credentials["aws_access_key_id"],
-                    # "secretKey": credentials["aws_secret_access_key"],
                     "assumeRoleArn": self.stack["TowerRoleArn"],
                 },
                 "description": f"Credentials for {self.stack_name}",
--- a/bin/configure-tower-projects.py
+++ b/bin/configure-tower-projects.py
@@ -536,7 +536,7 @@ class TowerWorkspace:
                 "provider": "aws",
                 "keys": {
                     "accessKey": credentials["aws_access_key_id"],
-                    #"secretKey": credentials["aws_secret_access_key"],
+                    # "secretKey": credentials["aws_secret_access_key"],
                     "assumeRoleArn": self.stack["TowerRoleArn"],
                 },
                 "description": f"Credentials for {self.stack_name}",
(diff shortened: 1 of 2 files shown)


thomasyu888 (commented):
🔥 Great work here. I do have a couple of questions around overall deployment and impact, but Ill defer to @BWMac first for a review.

danlu1:
@BWMac @thomasyu888 There might be another round of testing since the InstanceProfile has not been generated as this [instruction](https://docs.seqera.io/platform/23.4.0/enterprise/advanced-topics/use-iam-role) for the TowerRole. I'd like to test and see if we can bypass it.

bin/configure-tower-projects.py
```diff
@@ -528,16 +528,16 @@ def create_credentials(self) -> int:
...
         credentials = self.org.aws.get_secret_value(secret_arn)
         data = {
             "credentials": {
                 "name": self.stack_name,
                 "provider": "aws",
                 "keys": {
                     "accessKey": credentials["aws_access_key_id"],
-                    "secretKey": credentials["aws_secret_access_key"],
-                    "assumeRoleArn": self.stack["TowerForgeServiceRoleArn"],
+                    # "secretKey": credentials["aws_secret_access_key"],
```
BWMac: Should this line just be removed?

bin/configure-tower-projects.py
```diff
@@ -528,16 +528,16 @@ def create_credentials(self) -> int:
                 assert cred["deleted"] is None
                 return cred["id"]
         # Otherwise, create a new credentials entry for the project
-        secret_arn = self.stack["TowerForgeServiceUserAccessKeySecretArn"]
+        secret_arn = self.stack["TowerRoleArn"]
```
BWMac, line 531: So now `TowerRoleArn` serves the purpose of both `TowerForgeServiceUserAccessKeySecretArn` and `TowerForgeServiceRoleArn`?

danlu1: Yup. This is my understanding after reviewing Seqera instruction. We would be able to use IAM role arn rather than IAM user's secrets generated for the user to authenticate AWS.

bin/configure-tower-projects.py
danlu1: Will do.

BWMac (approved):
Looks good to me!

PR #312, opened 2024-06-11
[IT-3968] remove iam users

danlu1:
Problem:

The AWS credentials can't be populated with IAM role arn. And `get_secret_value` function does not work with IAM role arn.  Ideally, after converting to using IAM role, we no longer need to use accessKeys as the AWS credential of the Tower workspaces. Instead, we use AWS arn (see this [instruction](https://docs.seqera.io/platform/23.4.0/enterprise/advanced-topics/use-iam-role#:~:text=Log%20in%20to%20Seqera%20and%20create%20a%20new%20AWS%20credential.%20You%20are%20now%20prompted%20for%20an%20AWS%20arn%20instead%20of%20access%20keys.))

Solution: 

As the [Seqera's instruction](https://docs.seqera.io/platform/23.4.0/enterprise/advanced-topics/use-iam-role#:~:text=Create%20an%20instance%20profile%3A), an instance profile should be created and attached with the IAM role. My understanding is that creating InstanceProfile for the role would allow EC2 to receive the IAM role credential when it assumes the role. 

Testing:
The credential can be saved in Tower right now, but another error: {'message': "Cannot determine region of bucket 'example-dev-project-tower-scratch'"} popped up when generating the computation environment. The errors lie in these [lines](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/ee144038319caf89e095c6cf6ca235aa1aa1540e/bin/configure-tower-projects.py#L725C1-L730C1)

### DIFF
--- a/bin/configure-tower-projects.py
+++ b/bin/configure-tower-projects.py
@@ -528,14 +528,14 @@ class TowerWorkspace:
                 assert cred["deleted"] is None
                 return cred["id"]
         # Otherwise, create a new credentials entry for the project
-        secret_arn = self.stack["TowerRoleArn"]
-        credentials = self.org.aws.get_secret_value(secret_arn)
+        # secret_arn = self.stack["TowerRoleArn"]
+        # credentials = self.org.aws.get_secret_value(secret_arn)
         data = {
             "credentials": {
                 "name": self.stack_name,
                 "provider": "aws",
                 "keys": {
-                    "accessKey": credentials["aws_access_key_id"],
+                    # "accessKey": credentials["aws_access_key_id"],
                     "assumeRoleArn": self.stack["TowerRoleArn"],
                 },
                 "description": f"Credentials for {self.stack_name}",
--- a/templates/tower-project.j2
+++ b/templates/tower-project.j2
@@ -173,6 +173,11 @@ Resources:
               AWS:
                 - 'Fn::ImportValue': !Sub ${AWS::Region}-nextflow-ecs-service-shared-EcsServiceRoleArn
             Action: sts:AssumeRole
+  TowerRoleProfile:
+    Type: AWS::IAM::InstanceProfile
+    Properties:
+      Roles:
+        - !Ref TowerRole
 
   TowerForgeBatchHeadJobPolicy:
     Type: AWS::IAM::Policy


zaro0508 (commented):
@danlu1 can you please explain in this PRs description why you have taken this approach to fixing the problem and why you think this will fix it?

thomasyu888:
> @danlu1 can you please explain in this PRs description why you have taken this approach to fixing the problem and why you think this will fix it?

@zaro0508 we actually aren't sure what is going on.  @BWMac / @danlu1 can you fill @zaro0508 in?  Maybe he can also help us!

danlu1:
@zaro0508 I would appreciate your insights on these changes (this PR along with its prior [PR](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/pull/311)). 

Based on my preliminary research, the IAM role for Seqera, named `TowerRole`, has been added in your [PR](https://github.com/zaro0508/nextflow-infra/blob/6fbfefafa0a17a5d42112e3f986868b3246c7b5b/templates/tower-project.j2#L144C1-L176C1). The required IAM role policy ([nextflow-launch-iam-policy](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/dev/templates/nextflow-launch-iam-policy.yaml) and [nextflow-forge-iam-policy](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/dev/templates/nextflow-forge-iam-policy.yaml)) and trust policy (in the [AssumeRolePolicyDocument](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/80dd5587e32a788a22fe44027ea2cfc5dd9fe7f9/templates/tower-project.j2#L150) component of the TowerRole) are in place. 

As Seqera's instruction regarding creating [IAM artifacts](https://docs.seqera.io/platform/23.2.0/enterprise/advanced-topics/use-iam-role#create-the-iam-artifacts:~:text=Create%20the%20IAM%20artifacts%E2%80%8B,Instance%20%2D%2Drole%2Dname%20Seqera%2DRole), I submitted this [PR](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/pull/312) by adding a [TowerRoleProfile](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/ee2d05237b127a223dc2bab815cfcc3fea16a760/templates/tower-project.j2#L176) and attach the `TowerRole` to it. After creating the instance profile,  EC2 should be able to receive the IAM role credential when it assumes the `TowerRole`. I tested this PR on my local and successfully added the aws credential in the [Tower-dev project.](https://tower-dev.sagebionetworks.org/orgs/Sage-Bionetworks/workspaces/example-dev-project/credentials/6c5GT4SoKedFpSYIpc0dVm) 

Even though the credential has been added, another error: ```{'message': "Cannot determine region of bucket 'example-dev-project-tower-scratch'"}```(the bucket is the `workDir` of the Tower project) occurs when I tried to generate the computation environment for tower using the [configure-tower-projects.py](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/dev/bin/configure-tower-projects.py) , and the errors lie in these [lines](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/ee144038319caf89e095c6cf6ca235aa1aa1540e/bin/configure-tower-projects.py#L725C1-L730C1). This confused me a lot since `TowerRole` has been added to [Bucket policy](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/ee2d05237b127a223dc2bab815cfcc3fea16a760/templates/tower-project.j2#L383) and has GetBucketLocation permission.

Can you enlighten me on what could cause the issue? Also, can you share more on how you expect the `EcsServiceRole` to come into play? Thanks in advance.

bin/configure-tower-projects.py
```diff
@@ -528,14 +528,14 @@ def create_credentials(self) -> int:
...
-        credentials = self.org.aws.get_secret_value(secret_arn)
+        # secret_arn = self.stack["TowerRoleArn"]
+        # credentials = self.org.aws.get_secret_value(secret_arn)
         data = {
             "credentials": {
                 "name": self.stack_name,
                 "provider": "aws",
                 "keys": {
-                    "accessKey": credentials["aws_access_key_id"],
+                    # "accessKey": credentials["aws_access_key_id"],
```
zaro0508, line 538: i believe the key to setting this `key` is to look at line #544, `response = self.tower.request("POST", endpoint, params=params, json=data)`.  That is using the [seqera tower credentials API](https://cloud.seqera.io/openapi/index.html?_gl=1*yd5xxw*_gcl_au*MTIxMzIyMTA3My4xNzMwNDkyODI0#post-/credentials) to `Creates new Tower credentials`.  According to the API docs, passing in an `assumeRoleArn` also requires `accessKey` and `secretKey` therefore this will not work.

zaro0508 (changes requested):

bin/configure-tower-projects.py
danlu1: Yes, this is the confusing part I have. I remembered it worked when keeping the `accessKey` and `secretKey`. However, as my understanding based on this [doc](https://docs.seqera.io/platform/23.4/enterprise/advanced-topics/use-iam-role#:~:text=create%20a%20new%20AWS%20credential.%20You%20are%20now%20prompted%20for%20an%20AWS%20arn%20instead%20of%20access%20keys.), the `accessKey` and `secretKey` should be no longer needed if implementing role based authentication. Instead, the AWS role arn is used after the IAM artifacts is created and new credentials are added in Seqera.

zaro0508: Unfortunately this deployment setup does not fit with the assume role use case that's explained in the seqera docs.  This script `configure-tower-projects.py` is [executed with a Github action](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/dev/.github/workflows/rw-deploy.yaml#L76-L80)  which means it is running in a Github supplied instance, NOT an AWS instance.  Only AWS instances are setup to assume roles that give access to the Seqera app (and APIs).

Also the doc you referenced says `AWS-based customers can configure Seqera Platform to authenticate to AWS services like Batch with an IAM Role instead of IAM user credentials` which means that setup allows Seqera access to access AWS.  However this script is doing a POST request to Seqera which means it needs access to Seqera.  There is no way to allow a Github instances to assume a role to access Seqera unless Seqera allows setting up a trust relationship (via OIDC) between the two entities.

PR #321, opened 2024-06-20
Fixes Redis Config

BWMac:
The last deployment [failed](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/actions/runs/9604259386/job/26489377669) due to an issue with the `RedisContainerImage` configuration. 

We haven't been using the steps involving redis in our deployment with this [variable](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/1787a9e4d970b4871db31937cd6b2c0c5d06ea10/config/infra-dev/nextflow-ecs-task-definition.yaml#L42) being set to `false`.

I added the Seqera-hosted `redis` image as `RedisContainerImage` in case we need to enable it. Right now it is unclear to me if it needs to be enabled or not.

I also removed the credential configuration for `RedisContainerImage` in the ECS template because it is a public container (this is what was causing the error - a reference to `RedisContainerImage` when it wasn't defined.


thomasyu888 (approved):
🔥 LGTM!

PR #325, opened 2024-06-21
remove CronContainer from ContainerDefinitions  config

BWMac:
This PR removes `CronContainer` from the `ContainerDefinitions` config. This is because it uses the same container image as the backend and the configurations [must be unique](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/actions/runs/9617403116/job/26529103452#step:9:106).


templates/nextflow-ecs-task-definition.j2
```diff
@@ -201,10 +201,6 @@ Resources:
           Image: !Ref BackendContainerImage
           RepositoryCredentials:
             CredentialsParameter: !Sub 'arn:aws:secretsmanager:us-east-1:${AWS::AccountId}:secret:TOWER_DEV_SEQERA_REGISTRY_SECRET'
-        - Name: !Ref CronContainerName
```
thomasyu888, line 204: Should we try unique names as well later? 

!Sub ${BackendContainerName}-secret?

thomasyu888 (approved):
🔥 LGTM!

PR #326, opened 2024-06-21
Removes Duplicate `ContainerDefinitions`

BWMac:
The last deployment attempt threw an [error](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/actions/runs/9618168340/job/26533751927) about duplicate container definition names and I realized that we added additional blocks instead of providing the `RepositoryCredentials` to the existing container configs.


thomasyu888 (approved):
🔥 LGTM!

PR #329, opened 2024-07-02
[IBCDPE-1000] Adds EcsTaskExecutionRole to task definition template

BWMac:
This PR adds the EcsTaskExecutionRole configuration to the task-definition template so that the creation of the role is no longer manual.


BryanFauble (approved):
LGTM!

PR #336, opened 2024-07-10
Create model-ad-rna-project.yaml

jaclynbeck-sage:
Creates a project for running RNAseq data processing for Model-AD, on STRIDES.

### DIFF
--- a/config/projects-ampad/model-ad-rna-project.yaml
+++ b/config/projects-ampad/model-ad-rna-project.yaml
@@ -1,0 +1,25 @@
+template:
+  path: tower-project.j2
+stack_name: model-ad-rna-project
+dependencies:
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
🔥 LGTM!  We are in the midst of doing upgrades, so this may take a couple of weeks to get in. Is that ok?

jaclynbeck-sage:
> 🔥 LGTM! We are in the midst of doing upgrades, so this may take a couple of weeks to get in. Is that ok?

@thomasyu888 That is not ideal :( I was hoping to have access within a few days. But if that's not possible/easy to do, I do have access to someone else's project on STRIDES so I can run stuff there and replicate the pipeline once this project is live.

thomasyu888:
Got it @jaclynbeck-sage , let me try and get this through.

PR #344, opened 2024-07-17
[IBCDPE-1017] Adds Vesteinn Thorsson (ISB) to `iatlas-project`

BWMac:
This PR adds access to the `iatlas-project` workspace for Vesteinn Thorsson (PI at ISB) and removes access for Clarisse Lau (also ISB).

### DIFF
--- a/templates/nextflow-tower-config-bucket.yaml
+++ b/templates/nextflow-tower-config-bucket.yaml
@@ -68,7 +68,6 @@ Resources:
                 - "*@sagebase.org"
                 - "*@sagebionetworks.org"
                 - "cheimann@isbscience.org"
-                - "clarisse.lau@isbscience.org"
                 - "vesteinn.thorsson@isbscience.org"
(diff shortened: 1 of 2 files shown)


templates/nextflow-tower-config-bucket.yaml
```diff
@@ -69,6 +69,7 @@ Resources:
                 - "*@sagebionetworks.org"
                 - "cheimann@isbscience.org"
                 - "clarisse.lau@isbscience.org"
```
thomasyu888: Can we remove Clarisse as well? Not sure if it should be a separate PR.

thomasyu888 (approved):
🔥 Pre-approving! Thank you!

templates/nextflow-tower-config-bucket.yaml
BWMac: I think it's fine if we do it in one

PR #345, opened 2024-07-17
[IBCDPE-839] Add bryan.fauble@sagebase.org to iatlas, and pegs-challenge

BryanFauble:
**Problem:**

1. I am working to create a user guide that explains how to control nextflow execution costs, but I am not a part of any actively used projects. I would like to be added to some active projects so I can run some profiling scripts against previous workflow runs.


BryanFauble:
@BWMac , @jaymedina Since you two do a lot of work in this space, are there other projects you might suggest that may be better to look at? My goal is to use this script: https://github.com/Sage-Bionetworks-Workflows/py-optimize-nextflow - To look at previous workflow runs.

jaymedina:
Cool stuff! I know @BWMac has done work for the Dynamic Challenge, which sounds like a good candidate for this analysis because it's a forever ongoing Challenge (to my knowledge).

thomasyu888 (approved):
🔥 LGTM!

PR #354, opened 2024-09-12
[IT-3895] comment out access for Rixing/Phil to avoid 1024 limit

thomasyu888:
**Problem**
Running into character limit for S3 ARNs

**Solution**
Reduce personnel that have permissions to the project


thomasyu888:
Before merging, ask Phil/Rixing if they are running any automated jobs in HTAN project

zaro0508 (approved):
LGTM, another possible solution would be to try @jesusaurus suggestion of removing the `S3ReadWriteAccessArns` output because there doesn't seem to be anything that  depends on it.

thomasyu888:
> LGTM, another possible solution would be to try @jesusaurus suggestion of removing the `S3ReadWriteAccessArns` output because there doesn't seem to be anything that depends on it.

Thanks!  this error has occurred one other time, if it happens again, I'll remember this solution

PR #355, opened 2024-10-14
[IT-3929] Upgrade MySql version

zaro0508:
MySql 5.7 is EOL therefore we need to upgrade to the new AWS supported version of MySql

This is going to do an `In-place upgrade`[1].  The idea is to upgrade it in dev environment -> test -> propogate to prod -> test.

[1] https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Aurora.MySQL57.EOL.html#Aurora-Performing-an-Upgrade

### DIFF
--- a/templates/nextflow-aurora-mysql.yaml
+++ b/templates/nextflow-aurora-mysql.yaml
@@ -63,9 +63,9 @@ Parameters:
 Mappings:
   ClusterSettings:
     mysql:
-      version: 5.7.mysql_aurora.2.07.1
+      version: 8.0.mysql_aurora.3.07.1
       engine: aurora-mysql
-      family: aurora-mysql5.7
+      family: aurora-mysql8.0
 
 Conditions:
   UseSnapshotTrue: !Not [!Equals [!Ref SnapshotName, '']]


BryanFauble (approved):
Thanks for the update! Are there any breaking changes that your aware of?

thomasyu888:
We don't have a staging, but it'd be nice if this: https://sagebionetworks.jira.com/browse/IT-3888 was in place to automate the testing of this.  We can still probably merge and wait for Brad to be back from Vacation to fully test dev deployment

thomasyu888 (approved):

zaro0508:
> Thanks for the update! Are there any breaking changes that your aware of?

I don't know much about the software that's using the DB so i kinda have no idea.  I'm guessing @BWMac would probably know and be testing that it works after the upgrade.

BWMac (approved):
LGTM!

In their docs, Seqera states that the new version is supported and they do not give any warnings that I am aware of in terms of breaking changes so 🤞🏻 it plays out that way in our deployment.

zaro0508:
Manually tested this but in-place replacement didn't work.  
AWS returned following error.. 
```
UPDATE_FAILED Resource handler returned message: 
"Cannot find upgrade target from 5.7.serverless_mysql_aurora.2.11.4 
with requested version 8.0.mysql_aurora.3.07.1 for serverless 
engine mode
```

We will need to do a full migration.

PR #359, opened 2024-11-20
Create jhu-ras-pathway-project

ajs3nj:


thomasyu888:
@BWMac we may be at max compute envs right?

BWMac:
@thomasyu888 The Batch dashboard shows we are at 47 compute environments so I think we _should_ be ok to add one more and then we would be at the max.

image

thomasyu888:
> @thomasyu888 The Batch dashboard shows we are at 47 compute environments so I think we _should_ be ok to add one more and then we would be at the max.
> 
> image

Thanks @BWMac , feel free to review then we can add!

That's a bit scary, let me talk to our AWS reps to see if we can just bump the batch limit higher...

BWMac:
@thomasyu888 case to bump up priority on [IT-2361](https://sagebionetworks.jira.com/browse/IT-2361)?

[IT-2361]: https://sagebionetworks.jira.com/browse/IT-2361?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

BWMac (approved):

PR #381, opened 2025-03-11
[IT-3929] Update MySQL DB cluster

zaro0508:
Attempts to do in-place MySql DB upgrade from v5.7 to v8.0 does not work therefore this is going to do a blue/green upgrade. We will create a new parallel MySql v8.0 DB cluster, migrate the existing data from the old DB to the new DB then switch the Seqera app servers to use the new DB.

### DIFF
--- a/config/infra-dev/nextflow-aurora-mysql8.yaml
+++ b/config/infra-dev/nextflow-aurora-mysql8.yaml
@@ -1,0 +1,26 @@
+# MySql cluster version 3.0 with serverless v2
+template:
+  path: nextflow-aurora-mysql.yaml
+stack_name: nextflow-aurora-mysql8
+dependencies:
+  - infra-dev/workflows-kms-key.yaml
+  - infra-dev/nextflow-vpc.yaml
+  - infra-dev/nextflow-ecs-security-group.yaml
+
+parameters:
+  DBClusterIdentifier: tower8
+  DBName: tower8
+  DbClusterMapping: mysql8
+  KmsKeyAlias: alias/nextflow-aurora-bootstrap-lambda8
+  VpcID: !stack_output_external nextflow-vpc::VPCId
+  SubnetIDs:
+    - !stack_output_external nextflow-vpc::PrivateSubnet1
+    - !stack_output_external nextflow-vpc::PrivateSubnet2
+  EcsSecurityGroupId: !stack_output_external nextflow-ecs-security-group::SecurityGroupId
+  TemplateRootUrl: 'https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com'
+  AccountAdminArns:
+    - {{stack_group_config.sso_admin_role.arn}}
+    - !stack_output_external sagebase-github-oidc-workflows-dev-nextflow-infra::ProviderRoleArn
+
+stack_tags:
+  {{stack_group_config.default_stack_tags}}
--- a/templates/nextflow-aurora-mysql.yaml
+++ b/templates/nextflow-aurora-mysql.yaml
@@ -79,12 +79,14 @@ Mappings:
       version: 5.7.mysql_aurora.2.07.1
       engine: aurora-mysql
       family: aurora-mysql5.7
-      enginemode: serverless
+      EngineMode: serverless
+      EnableHttpEndpoint: true
     mysql8:
       version: 8.0.mysql_aurora.3.04.0
       engine: aurora-mysql
       family: aurora-mysql8.0
-      enginemode: provisioned
+      EngineMode: provisioned
+      EnableHttpEndpoint: false
 
 Conditions:
   UseSnapshotTrue: !Not [!Equals [!Ref SnapshotName, '']]
@@ -240,7 +242,7 @@ Resources:
       - CloudWatchDBClusterSlowQueryLogGroup
     Properties:
       Engine: !FindInMap [ ClusterSettings, !Ref DbClusterMapping, engine ]
-      EngineMode: !FindInMap [ ClusterSettings, !Ref DbClusterMapping, enginemode ]
+      EngineMode: !FindInMap [ ClusterSettings, !Ref DbClusterMapping, EngineMode ]
       EngineVersion: !FindInMap [ ClusterSettings, !Ref DbClusterMapping, version ]
       DatabaseName: !Ref DBName
       DBClusterIdentifier: !Ref DBClusterIdentifier
@@ -256,7 +258,7 @@ Resources:
       StorageEncrypted:  !If [UseSnapshotTrue, !Ref 'AWS::NoValue', true]
       DBClusterParameterGroupName: !Ref AuroraClusterParameterGroup
       BackupRetentionPeriod: 14
-      EnableHttpEndpoint: true
+      EnableHttpEndpoint: !FindInMap [ ClusterSettings, !Ref DbClusterMapping, EnableHttpEndpoint ]
       DBSubnetGroupName: !Ref DBSubnetGroup
       VpcSecurityGroupIds:
         - !Ref ClusterSecurityGroup
--- a/templates/nextflow-aurora-mysql.yaml
+++ b/templates/nextflow-aurora-mysql.yaml
@@ -60,12 +60,31 @@ Parameters:
     Type: List<String>
     Description: List of admin IAM user and role ARNs (strings)
 
+  DbClusterMapping:
+    Type: String
+    Description: List of admin IAM user and role ARNs (strings)
+    AllowedValues:
+      - mysql5
+      - mysql8
+    Default: mysql5
+
+  KmsKeyAlias:
+    Type: String
+    Description: Alias name of KMS key to use for encryption
+    Default: alias/nextflow-aurora-bootstrap-lambda
+
 Mappings:
   ClusterSettings:
-    mysql:
+    mysql5:
       version: 5.7.mysql_aurora.2.07.1
       engine: aurora-mysql
       family: aurora-mysql5.7
+      enginemode: serverless
+    mysql8:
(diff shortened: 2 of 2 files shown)


templates/nextflow-aurora-mysql.yaml
```diff
@@ -291,7 +310,7 @@ Resources:
       Description: Bootstrap newly Created Aurora Database
       Handler: index.lambda_handler
       Role: !GetAtt BootstrapLambdaRole.Arn
-      Runtime: python3.7
+      Runtime: python3.10
```
zaro0508, line 333: python3.7 is EOL, it won't even deploy anymore.  Need to update to never version. https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

templates/nextflow-aurora-mysql.yaml
```diff
@@ -60,12 +60,31 @@ Parameters:
...
+    mysql5:
       version: 5.7.mysql_aurora.2.07.1
       engine: aurora-mysql
       family: aurora-mysql5.7
+      enginemode: serverless
+    mysql8:
+      version: 8.0.mysql_aurora.3.04.0
+      engine: aurora-mysql
+      family: aurora-mysql8.0
+      enginemode: provisioned
```
zaro0508: from the docs: The serverless engine mode only applies for Aurora Serverless v1 DB clusters. Aurora [Serverless v2 DB clusters use the provisioned engine mode](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbcluster.html#cfn-rds-dbcluster-enginemode).

templates/nextflow-aurora-mysql.yaml
```diff
@@ -60,12 +60,31 @@ Parameters:
...
 Mappings:
   ClusterSettings:
-    mysql:
+    mysql5:
       version: 5.7.mysql_aurora.2.07.1
       engine: aurora-mysql
       family: aurora-mysql5.7
+      enginemode: serverless
+    mysql8:
+      version: 8.0.mysql_aurora.3.04.0
```
zaro0508: [Aurora v5.7 maps to Serverless 1 & MySql 2.x and Aurora v8.0 maps to Serverless 2 & MySql 3.x](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbcluster.html#cfn-rds-dbcluster-engineversion)

[Serverless 2 maps to MySql 3.x](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Concepts.Aurora_Fea_Regions_DB-eng.Feature.ServerlessV2.html)

config/infra-dev/nextflow-aurora-mysql8.yaml
```diff
@@ -0,0 +1,26 @@
...
+dependencies:
+  - infra-dev/workflows-kms-key.yaml
+  - infra-dev/nextflow-vpc.yaml
+  - infra-dev/nextflow-ecs-security-group.yaml
+
+parameters:
+  DBClusterIdentifier: tower8
+  DBName: tower8
+  DbClusterMapping: mysql8
+  KmsKeyAlias: alias/nextflow-aurora-bootstrap-lambda8
```
zaro0508, line 14: Provide new resource names to avoid conflicting with existing names

templates/nextflow-aurora-mysql.yaml
```diff
@@ -60,12 +60,33 @@ Parameters:
...
       version: 5.7.mysql_aurora.2.07.1
       engine: aurora-mysql
       family: aurora-mysql5.7
+      EngineMode: serverless
+      EnableHttpEndpoint: true
+    mysql8:
+      version: 8.0.mysql_aurora.3.04.0
+      engine: aurora-mysql
+      family: aurora-mysql8.0
+      EngineMode: provisioned
```
zaro0508, line 88: from the docs: The serverless engine mode only applies for Aurora Serverless v1 DB clusters. Aurora [Serverless v2 DB clusters use the provisioned engine mode](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbcluster.html#cfn-rds-dbcluster-enginemode).

templates/nextflow-aurora-mysql.yaml
```diff
@@ -60,12 +60,33 @@ Parameters:
...
+      EnableHttpEndpoint: false
```
zaro0508: Get the following error when set to `false`:
```
infra-dev/nextflow-aurora-mysql8 AuroraCluster AWS::RDS::DBCluster CREATE_FAILED
Resource handler returned message: "The engine version that you specified doesn't
support the HTTP endpoint for using RDS Data API.
```
[Aurora MySQL Serverless v2 does not yet support data API. Therefore, EnableHttpEndpoint cannot be enabled](https://repost.aws/questions/QUGYgbABjWRXaSC1mRH_OWFQ/error-while-enabling-http-endpoint-for-rds-serverless-version-2)


Get the following error when set to `true`:
```
infra-dev/nextflow-aurora-mysql8 BootstrapLambdaTrigger
Custom::BootstrapLambdaTrigger CREATE_FAILED Received response status [FAILED]
from custom resource. Message returned: An error occurred (HttpEndpointNotEnabledException)
when calling the ExecuteStatement operation: HttpEndpoint is not enabled for resource
arn:aws:rds:us-east-1:035458030717:cluster:tower8. 
```

***The `BootstrapLambdaFunction` resource in this file requires setting `EnableHttpEndpoint: true` however `true` is not supported for Serverless 2 which is mapped to v8.0.***

thomasyu888 (approved):
LGTM! Thanks!

BWMac (approved):
Thanks @zaro0508

templates/nextflow-aurora-mysql.yaml
zaro0508: > EnableHttpEndpoint: true however true is not supported for Serverless 2 which is mapped to v8.0.

[Data API supports Aurora MySQL 3.07 and higher versions](https://aws.amazon.com/about-aws/whats-new/2024/09/amazon-aurora-mysql-rds-data-api).  Just need to deploy a newer version.

templates/nextflow-aurora-mysql.yaml
```diff
@@ -112,7 +133,7 @@ Resources:
       GenerateSecretString:
         SecretStringTemplate: !Sub '{"username": "${DBUserName}"}'
         GenerateStringKey: password
-        ExcludeCharacters: '"@/\$`&'
+        ExcludePunctuation: true
```
zaro0508, line 136: The previous use of `ExcludeCharacters` property would cause intermittent failures because it did not exclude all punctuations.  For example, on occasion this will generate passwords with a single quote character which would cause failures in the SQL syntax in the `BootstrapLambdaFunction` because single quotes needs to be escaped.

templates/nextflow-aurora-mysql.yaml
```diff
@@ -237,15 +258,33 @@ Resources:
...
       VpcSecurityGroupIds:
         - !Ref ClusterSecurityGroup
 
+  DBParameterGroup:
+    Type: 'AWS::RDS::DBParameterGroup'
+    Properties:
+      Description: Aurora Database Parameter Group
+      Family: aurora-mysql8.0
+
+  DBInstance1:
```
zaro0508, line 272: We need to explicitly create a db instance because aurora-mysql v8 can contain mulitple db instances in a cluster.  https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Aurora.CreateInstance.html

templates/nextflow-aurora-mysql.yaml
```diff
@@ -310,11 +349,11 @@ Resources:
           log.setLevel(logging.DEBUG)
 
           def submit_sql(sql):
+            log.debug(f'Submitting SQL: {sql}')
             return boto3.client('rds-data').execute_statement(
               secretArn=environ['MASTER_SECRET_ARN'],
               database=environ['DB_NAME'],
               resourceArn=environ['CLUSTER_ARN'],
-              schema='mysql',
```
zaro0508, line 317: the boto3 rds-data client [no longer supports schema](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds-data/client/execute_statement.html).

templates/nextflow-aurora-mysql.yaml
```diff
@@ -326,9 +365,11 @@ Resources:
               secret = json.loads(secret_response['SecretString'])
               username = secret['username']
               password = secret['password']
+              log.debug(f'submit to database: {environ["DB_NAME"]}')
+              log.debug(f'using resourceArn: {environ["CLUSTER_ARN"]}')
               alter_db_response = submit_sql(f"ALTER DATABASE {db_name} CHARACTER SET utf8 COLLATE utf8_bin;")
               log.debug(f"Alter DB response: {alter_db_response}")
-              create_user_response = submit_sql(f"CREATE USER '{username}' IDENTIFIED BY '{password}';")
+              create_user_response = submit_sql(f"CREATE USER IF NOT EXISTS '{username}' IDENTIFIED BY '{password}';")
```
zaro0508, line 372: When deployed, this event would occur multiple times and cause this statement to fail because the user might already have been created with a previous event.  AWS says that duplicate events are common and [users need to handle that by making their lambdas idempotent](https://repost.aws/knowledge-center/lambda-function-idempotent).  This change makes it work if we get duplicate events to run this script

zaro0508:
The template needed for aurora mysql v8 is different enough from 5.7 that it would be difficult to support both versions in a single templates.  I've decided to abandon this PR for #383 which creates a separate template for aurora mysql v8.  Once we complete migrating to v8 we can just delete the templates for v5.7.

PR #388, opened 2025-03-18
[IT-3929] Update MySQL DB cluster

zaro0508:
Attempts to do in-place MySql DB upgrade from v5.7 to v8.0 does not work therefore we are going to do a blue/green upgrade. We will create a new parallel MySql v8.0 DB cluster, manually migrate the existing data from the old 5.7 DB to the new 8.0 DB then switch the Seqera app servers to use the new DB.

this is a redo of PR #381 with reason https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/pull/381#issuecomment-2734547689

### DIFF
--- a/config/infra-dev/nextflow-aurora-mysql8.yaml
+++ b/config/infra-dev/nextflow-aurora-mysql8.yaml
@@ -1,0 +1,23 @@
+template:
+  path: nextflow-aurora-mysql8.yaml
+stack_name: nextflow-aurora-mysql8
+dependencies:
+  - infra-dev/workflows-kms-key.yaml
+  - infra-dev/nextflow-vpc.yaml
+  - infra-dev/nextflow-ecs-security-group.yaml
+
+parameters:
+  DBClusterIdentifier: tower8
+  DBName: tower8
+  VpcID: !stack_output_external nextflow-vpc::VPCId
+  SubnetIDs:
+    - !stack_output_external nextflow-vpc::PrivateSubnet1
+    - !stack_output_external nextflow-vpc::PrivateSubnet2
+  EcsSecurityGroupId: !stack_output_external nextflow-ecs-security-group::SecurityGroupId
+  TemplateRootUrl: 'https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com'
+  AccountAdminArns:
+    - {{stack_group_config.sso_admin_role.arn}}
+    - !stack_output_external sagebase-github-oidc-workflows-dev-nextflow-infra::ProviderRoleArn
+
+stack_tags:
+  {{stack_group_config.default_stack_tags}}
--- a/templates/nextflow-aurora-mysql8.yaml
+++ b/templates/nextflow-aurora-mysql8.yaml
@@ -1,0 +1,461 @@
+AWSTemplateFormatVersion: 2010-09-09
+Description: Serverless Aurora MySQL DB for NextFlow Tower
+
+Parameters:
+
+  VpcID:
+    Description: ID of VPC
+    Type: AWS::EC2::VPC::Id
+
+  SubnetIDs:
+    Description: A list of subnets in the chosen VPC
+    Type: List<AWS::EC2::Subnet::Id>
+
+  DBClusterIdentifier:
+    Description: Unique identifier of databse cluster.
+    Type: String
+    MinLength: 1
+    MaxLength: 63
+    AllowedPattern: "^(?!.*--)[a-zA-Z]+[0-9a-zA-Z-]*[0-9a-zA-Z]$"
+    ConstraintDescription: >
+      1 to 63 alphanumeric characters or hyphens.
+      First character must be a letter.
+      Can't contain two consecutive hyphens.
+      Can't end with a hyphen.
+    Default: tower
+
+  DBName:
+    Description: Optional. Database Name within cluster.
+    Type: String
+    MinLength: 1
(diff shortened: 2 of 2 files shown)


config/infra-dev/nextflow-aurora-mysql8.yaml
```diff
@@ -0,0 +1,23 @@
...
+  path: nextflow-aurora-mysql8.yaml
+stack_name: nextflow-aurora-mysql8
+dependencies:
+  - infra-dev/workflows-kms-key.yaml
+  - infra-dev/nextflow-vpc.yaml
+  - infra-dev/nextflow-ecs-security-group.yaml
+
+parameters:
+  DBClusterIdentifier: tower8
+  DBName: tower8
```
zaro0508, line 11: Provide new resource names to avoid conflicting with existing names

templates/nextflow-aurora-mysql8.yaml
```diff
@@ -0,0 +1,461 @@
...
+    Type: List<String>
+    Description: List of admin IAM user and role ARNs (strings)
+
+Mappings:
+  ClusterSettings:
+    mysql:
+      version: 8.0.mysql_aurora.3.08.1
+      engine: aurora-mysql
+      family: aurora-mysql8.0
+      EngineMode: provisioned  #
```
zaro0508: from the docs: The serverless engine mode only applies for Aurora Serverless v1 DB clusters. [Aurora Serverless v2 DB clusters use the provisioned engine mode](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbcluster.html#cfn-rds-dbcluster-enginemode).

templates/nextflow-aurora-mysql8.yaml
```diff
@@ -0,0 +1,461 @@
...
+    Type: List<String>
+    Description: List of admin IAM user and role ARNs (strings)
+
+Mappings:
+  ClusterSettings:
+    mysql:
+      version: 8.0.mysql_aurora.3.08.1
+      engine: aurora-mysql
+      family: aurora-mysql8.0
+      EngineMode: provisioned
```
zaro0508, line 69: from the docs: The serverless engine mode only applies for Aurora Serverless v1 DB clusters. Aurora [Serverless v2 DB clusters use the provisioned engine mode](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbcluster.html#cfn-rds-dbcluster-enginemode).

templates/nextflow-aurora-mysql8.yaml
```diff
@@ -0,0 +1,461 @@
...
+            try:
+              db_name = environ['DB_NAME']
+              secret_response = boto3.client('secretsmanager').get_secret_value(SecretId=environ['DB_USER_SECRET_ARN'])
+              secret = json.loads(secret_response['SecretString'])
+              username = secret['username']
+              password = secret['password']
+              alter_db_response = submit_sql(f"ALTER DATABASE {db_name} CHARACTER SET utf8 COLLATE utf8_bin;")
+              log.debug(f"Alter DB response: {alter_db_response}")
+              create_user_response = submit_sql(f"CREATE USER IF NOT EXISTS '{username}' IDENTIFIED BY '{password}';")
+              log.debug(f"Create user response: {create_user_response}")
```
BryanFauble, line 351: Is it possible that this could leak the password if an exception was raised during the `return boto3.client('rds-data').execute_statement` call? Or if the SQL statement is returned in someway without an exception being raised?

templates/nextflow-aurora-mysql8.yaml
```diff
@@ -0,0 +1,461 @@
...
+    Description: Optional. Username to be created for database.
+    Type: String
+    MinLength: 1
+    MaxLength: 32
+    Default: tower
+
+  SnapshotName:
+    Description: Optional. Snapshot ID to restore database. Leave this blank if you are not restoring from a snapshot.
+    Type: String
+    Default: ''
```
BryanFauble, line 48: In the PR description you mention

> manually migrate the existing data from the old 5.7 DB to the new 8.0 DB then switch the Seqera app servers to use the new DB.

Is this `SnapshotName` where that manual data migration is going to happen, if it's possible to do a direct migration for a snapshot from 5.7 to 8.0?

BryanFauble (approved):
What I'm looking at here looks great to me!

BWMac (approved):

templates/nextflow-aurora-mysql8.yaml
zaro0508: No, it's not possible.  This was designed for migrating data from the same db version.  Going from mysql 5.7 to 8.0 will result in the following error.. `AuroraCluster AWS::RDS::DBCluster CREATE_IN_PROGRESS The engine version you requested for your restored DB cluster (8.0.mysql_aurora.3.08.1) is not compatible with the engine version of the DB cluster snapshot (5.7.mysql_aurora.2.11.4).`

templates/nextflow-aurora-mysql8.yaml
zaro0508: No, the error that is returned would be something like this.. `An error was encountered when bootstrapping: An error occurred (DatabaseErrorException) when calling the ExecuteStatement operation: Operation CREATE USER failed for 'tower'@'%'; Error code: 1396; SQLState: HY000`

PR #414, opened 2025-07-09
[IT-3929] New database version

zaro0508:
Create a new aurora MySQL v8.0 database for seqera tower. The template for the DB is based off of the original nextflow-aurora-mysql.yaml except that this DB will be used to migrate existing data onto it.  In this scenario we are creating a DB from an existing DB Snapshot which already contains the bootstrapping and data.  This scenario does not require a lambda, it assumes an existing DB snapshot to already exist.  We created the DB snapshot manually using instructions in issue IT-3929


thomasyu888 (approved):
LGTM!

PR #415, opened 2025-07-10
[IT-3929] switch to new database

zaro0508:
Switch nextflow from using the existing MySQL v5.7 to using the new MySQL v8.0 database for the dev environment.

depends on #414

PR #420, opened 2025-07-16
[IT-3929] Remove old tower database

zaro0508:
We have migrated to MySQL v8.0 database on dev environment. Now we remove the old MySQL v5.7 datbase on dev to make sure everthing works without it.

depends on PR #415


thomasyu888 (approved):

thomasyu888:
@zaro0508 NOTE, we haven't tested the dev deployment just yet but we will hopefully do that within this next month here, then we can find a time to do a migration

PR #423, opened 2025-07-22
[IT-4467] Add Nvidia based GPU instances

zaro0508:
According to the Seqera docs[1], to enable GPUs we need to add Nvidia GPUs and enable it in seqera compute environments.  Everything else required to use the GPUs is done with the Nextflow language[2] and nvidia container toolkit[3]

[1] https://docs.seqera.io/platform-cloud/compute-envs/overview#gpu-usage
[2] https://www.nextflow.io/docs/latest/process.html#accelerator
[3] https://github.com/NVIDIA/nvidia-container-toolkit

### DIFF
--- a/bin/configure-tower-projects.py
+++ b/bin/configure-tower-projects.py
@@ -52,6 +52,15 @@ NONGPU_EC2_INSTANCE_TYPES = (
     "m6i.8xlarge", "r5a.8xlarge", "r6a.8xlarge", "r6i.8xlarge"
 )
 
+GPU_EC2_INSTANCE_TYPES = (
+    # Amazon linux 2023 AMIs:
+    #   https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-optimized_AMI.html
+    # GPU instance types:
+    #   https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-gpu.html
+    "p3.2xlarge", "p3.8xlarge", "p3.16xlarge", "p3dn.24xlarge", "p4d.24xlarge",
+    "p5.48xlarge"
+)
+
 ECS_CONFIG = """
 ECS_CONTAINER_STOP_TIMEOUT=10m
 ECS_CONTAINER_START_TIMEOUT=10m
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
LGTM!

xschildw (approved):

bin/configure-tower-projects.py
```diff
@@ -52,6 +52,15 @@
     "m6i.8xlarge", "r5a.8xlarge", "r6a.8xlarge", "r6i.8xlarge"
 )
 
+GPU_EC2_INSTANCE_TYPES = (
+    # Amazon linux 2023 AMIs:
+    #   https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-optimized_AMI.html
+    # GPU instance types:
+    #   https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-gpu.html
+    "p4d.24xlarge", "p4de.24xlarge", "p5.48xlarge", "p5e.48xlarge",
+    "p5en.48xlarge", "p6-b200.48xlarge", "p6e-gb200.36xlarge"
```
vpchung, line 61: Thanks for adding these! Can we also include the cheaper GPU alternatives too? 😄 

```
"g5.xlarge", "g5.2xlarge", "g5.4xlarge", "g5.8xlarge", "g5.16xlarge"
```

vpchung (changes requested):

bin/configure-tower-projects.py
zaro0508: are you sure @vpchung ?  I thought `P` family is what you guys wanted.  Here's an good article about the difference between `P` and `G` families of GPUs. https://www.cloudoptimo.com/blog/aws-p-family-vs-g-family-a-detailed-comparison-for-high-performance-tasks/

vpchung: Hmm... `p` is definitely what we had used in the past (specifically, `p3` for the V100 GPU) but now that they are no longer supporting that, I hesitate to use `p4d.24xlarge` since it's so much more costly. 

But if that's not a factor, then yes, I think the A100 GPUs would be preferred over the A10Gs.

vpchung: After thinking on this more, I think you are right, @zaro0508 , let's stick with what you added.

vpchung (approved):

PR #427, opened 2025-08-12
[IT-3970] Upgrade Seqera Platform version

zaro0508:
Seqera data explorer is not working correctly due to a dabase schema mismatch.
Seqera support recommends upgrading Seqera to v25.2.0 to allow the app to
bootstrap the correct schema.

from support ticket #6590:
```
There appears to be a column (hidden) that your current version of
the Platform should not be referencing yet, as it's only introduced
in a later version. However, this column seems to be missing, which
is causing the issue with Data Explorer.
```


thomasyu888 (approved):
LGTM!

PR #431, opened 2025-08-13
Promote1

zaro0508:
Promote dev to prod

depends on #418


bin/configure-tower-projects.py
```diff
@@ -52,6 +52,15 @@
     "m6i.8xlarge", "r5a.8xlarge", "r6a.8xlarge", "r6i.8xlarge"
 )
 
+GPU_EC2_INSTANCE_TYPES = (
```
zaro0508, line 55: The CI cannot update compute environments, it can only create them therefore we need to delete the compute envs to allow the CI to create new ones with these new instance types.  Instead of updating all compute environments in every nextflow project we are going spot update by only removing the compute envs that need updating.  Which nextflow projects will need these new instance types @vpchung ?

vpchung: Thanks, @zaro0508 - there are currently no nextflow projects that are utilizing GPUs. So we're good 👍🏼

bin/configure-tower-projects.py
```diff
@@ -685,7 +694,7 @@ def generate_compute_environment(self, name: str, model: str) -> dict:
                         "ec2KeyPair": None,
                         "ecsConfig": ECS_CONFIG.strip(),
                         "efsCreate": False,
-                        "gpuEnabled": False,
+                        "gpuEnabled": True,
```
BryanFauble, line 697: **[No Action Needed]**

I was curious on what this flag meant, so I posted a question over to Github copilot when looking at this code, and this is what I got back out:



----

Looking at this code, You've identified an interesting configuration where `"gpuEnabled": True` is set while the `instance_types` list includes both GPU and non-GPU instances.

This configuration should **not** present an issue. Here's why:

## Purpose of the `gpuEnabled` flag

The `gpuEnabled` flag in AWS Batch compute environments serves as a capability declaration rather than a strict requirement. When set to `True`, it indicates that the compute environment is capable of handling GPU workloads, but it doesn't mandate that every job or instance must use GPUs.

## How it works in practice

1. **Instance Selection**: AWS Batch will intelligently select from the available instance types based on the job requirements. If a job requests GPU resources, Batch will choose from the GPU instance types (like `p4d.24xlarge`, `p5.48xlarge`, etc.). If a job doesn't require GPUs, it can still use the non-GPU instances.

2. **Flexibility**: This mixed approach provides maximum flexibility - the same compute environment can handle both GPU-intensive workloads (like machine learning training) and regular CPU-only tasks without needing separate compute environments.

3. **Resource Optimization**: AWS Batch's scheduler will make optimal choices based on:
   - Job resource requirements (CPU, memory, GPU)
   - Instance availability
   - Cost considerations (especially with the `SPOT_CAPACITY_OPTIMIZED` allocation strategy)

## Why this design makes sense

In the context of Nextflow workflows, this configuration is actually quite smart because:
- Some workflow steps may require GPU acceleration (e.g., deep learning inference)
- Other steps in the same workflow may only need CPU resources
- Having one compute environment that can handle both scenarios simplifies workflow management

So the `gpuEnabled: True` flag essentially means "this compute environment is GPU-capable" rather than "only use GPU instances."

BryanFauble (approved):
LGTM!

thomasyu888 (commented):
Thanks for this. Please Do not merge until I give a greenlight because there is still the database migration for prod, testing on dev, and development for a program to support the entry depreciation.

zaro0508:
replaced with PR #448

PR #432, opened 2025-08-15
Add Beatriz to model-ad-rna-project.yaml

jaclynbeck-sage:
I need to give Beatriz Saldana access to the Model-AD RNA seq project.


zaro0508 (approved):

thomasyu888:
@zaro0508 you may need to cherry pick this commit into production so it gets deployed as we are still undergoing testing and modification to support the new version of SP

PR #433, opened 2025-08-18
promote commit f8096d7de3 in dev to prod

zaro0508:
Cherry pick commit f8096d7de3 in PR #432 from dev to prod


config/projects-ampad/model-ad-rna-project.yaml
```diff
@@ -21,5 +22,5 @@ parameters:
 
 stack_tags:
   Department: NDR
-  Project: amp-ad
+  Project: model-ad-rna-project
```
ConsoleCatzirl, line 25: nit: I don't think the Project tag is used anywhere anymore, and this is missing a CostCenter tag

ConsoleCatzirl (approved):

config/projects-ampad/model-ad-rna-project.yaml
zaro0508: This was setup all before your work with cost center.  I believe CostCenter was left out in this case because ampad projects are deployed to the strides-ampad-workflow (751556145034) account which is completed paid by STRIDES.

PR #439, opened 2025-09-10
[IT-3929] Retain VPC resources on update

zaro0508:
Deployment thru the CI attempts to create new VPC resources and causing the following failure[1] even thought the cloudformation diff detection shows no changes.

```
infra-dev/nextflow-vpc NATGateway AWS::EC2::NatGateway UPDATE_FAILED Resource
handler returned message: "Error occurred during operation 'NatGateway
nat-099a46339af1c5fc9 is in state failed and hence failed to stabilize.
Detailed failure message: Network vpc-07290d198a10e8739 has no Internet
gateway attached'."
```

I am not able to reproduce locally, my attempt to update locally always returns no changes from cloudformation.

As a work around we explicitly tell clouformation to not create new resource when updating the template.
This is fine for a VPC because cloudformation has always had a difficult time updating VPCs.

[1] https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/actions/runs/17598318619/job/50007210182

### DIFF
--- a/templates/vpc-20-private.yaml
+++ b/templates/vpc-20-private.yaml
@@ -1,0 +1,364 @@
+# This template is originally from
+# https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.3.0/templates/VPC/vpc-20-private.yaml
+AWSTemplateFormatVersion: "2010-09-09"
+Metadata:
+  cfn-lint:
+    config:
+      ignore_checks: [W3011]
+Description: >
+  Creates a VPC of 4096 addresses, mostly in private subnets, and a public
+  subnet to accommodate a NAT gateway.
+Parameters:
+  VpcName:
+    Description: The VPC name (i.e. Synapse-Prod)
+    Type: String
+  VpcSubnetPrefix:
+    Description: The VPC subnet prefix (i.e. 10.255)
+    Type: String
+  PublicSubnetZones:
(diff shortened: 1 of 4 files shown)


templates/vpc-20-private.yaml
```diff
@@ -0,0 +1,364 @@
...
+      CIDR: "8.0/22"  # 10.255.8.1 - 10.255.11.254
+    Private3:
+      CIDR: "12.0/23" # 10.255.12.1 - 10.255.13.254
+    Public:
+      CIDR: "14.0/24" # 10.255.14.1 - 10.255.14.254
+    Public1:
+      CIDR: "15.0/24" # 10.255.15.1 - 10.255.15.254
+Resources:
+  VPC:
+    UpdateReplacePolicy: Retain
```
zaro0508, line 52: The only thing added from the [original template](https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.3.0/templates/VPC/vpc-20-private.yaml) are these [UpdateReplacePolicy: Retain](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-attribute-updatereplacepolicy.html) parameters.

thomasyu888 (approved):
LGTM! But i'd like to defer to an IT expert for final review. I'll defer to @zaro0508 on who this individual is unless you just want to merge and fix issues that arise.

PR #481, opened 2025-11-11
Revert "[DPESPRT-321] Add new users to tower viewer ARNs"

thomasyu888:
Reverts Sage-Bionetworks-Workflows/nextflow-infra#475, because this action failed: https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/actions/runs/19278844665


zaro0508 (changes requested):
You don't need to revert. The deployment to AWS workflow passed, it's the py-orca validation workflow that failed.  The failure was due to a missing service user we use for testing which was fixed in PR #478

PR #506, opened 2026-02-18
Fix integration tests

zaro0508:
the py-orca import "from orca.services.synapse import SynapseOps"
fails with "ModuleNotFoundError: No module named 'pkg_resources'".
This is due to a transitive dependency of setuptools version 82.0.0.
We need to pin to ver 81.0.0 for it to continue working.


thomasyu888 (approved):
LGTM

BryanFauble (approved):
Lgtm

PR #507, opened 2026-02-24
[IT-2361] Add a tower project to share CEs

zaro0508:
The purpose of this tower project is to allow other tower projects to reference this project's compute environments.  The idea is to conserve CE due to AWS 50 CE max quota. This will allow multiple tower projects to share this project's CEs.

### DIFF
--- a/config/projects-dev/shared-ce-project.yaml
+++ b/config/projects-dev/shared-ce-project.yaml
@@ -1,0 +1,24 @@
+template:
+  path: tower-project.j2
+stack_name: shared-ce-project
+dependencies:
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
Thanks @zaro0508 - one thing to look our for is to whether tagging of resources based on the different workspaces will work in this model.

That is, if I have. BDF workspace, will the AWS cost tags be transferred when using a shared AWS CE.

zaro0508:
> one thing to look our for is to whether tagging of resources based on the different workspaces will work in this model.

I have asked seqera support for clarification on how labels would work in this case.  We should also run some tests to see how it actually works.

PR #510, opened 2026-03-02
[DPESPRT-330] Remove user maria.diaz from YAML config

sophiajobe:
Removed maria.diaz@sagebase.org from the list of users for PEGS and Olfactory spaces


thomasyu888 (approved):
🔥 LGTM!

PR #518, opened 2026-03-09
[IT-4104] disable all Tower forge user access keys

zaro0508:
The AWS `last accessed` info indicates that these user keys are never accessed. I think seqera tower is using roles instead of user keys to access resources.  Disable all user keys to test my theory.


thomasyu888 (approved):
LGTM!

PR #521, opened 2026-03-10
[IT-2361] Set execution role for shared CE

zaro0508:
When Batch Forge creates a Compute Environment in Platform, it automatically configures the EC2 Instance role with an IAM PassRole permission scoped specifically to the Batch Execution Role created for that same compute environment.

To fix this, we need to configure the Batch Execution Role field on the Manual CE to use the execution role from shared CE project's Batch Forge CE's execution role.

- Refactor get_compute_env_queues to get_compute_env_config
  * Return full config dict instead of just queue tuple for flexibility
  * Allows accessing executionRole and other config fields
  * Simplifies error handling with single return statement

- Add check_existing_compute_env helper method
  * Validates platform (aws-batch) and status (AVAILABLE/CREATING)
 
- Improve create_manual_compute_environment
  * Extract executionRole from source CE config
  * Use check_existing_compute_env helper for consistency
  * Simplify control flow with single return statement
  * Standardize variable naming (comp_env_name -> compute_env_name)
  * Remove redundant conditional check before set_primary_compute_environment

- Fix get_compute_env_id_by_name logging
  * Move success message inside method for better visibility
  * Remove warning for not found case (handled by callers)

- Update comments for clarity
  * Clarify Batch Forge CE creation only when no manual CEs configured
  * Update docstrings to reflect execution role extraction

### DIFF
--- a/bin/configure-tower-projects.py
+++ b/bin/configure-tower-projects.py
@@ -420,7 +420,7 @@ class TowerWorkspace:
         self.populate()
         self.cleanup_compute_environments()
         if self.has_launchers():
-            # Check if manual compute environments are configured
+            # Only create Batch Forge CE if no manual CEs are configured
             if self.manual_compute_envs:
                 for manual_compute_env in self.manual_compute_envs:
                     source_workspace_name = manual_compute_env["WorkspaceName"]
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
LGTM!

PR #529, opened 2026-03-12
[IT-2361] Migrate CEs for 3 other projects

thomasyu888:
I confirmed with Robert and Aditi about migrating these resources and to the best of my knowledge, there is no jobs running in these workspaces as of yesterday

### DIFF
--- a/config/projects-prod/mc2-project.yaml
+++ b/config/projects-prod/mc2-project.yaml
@@ -53,6 +53,7 @@ template:
 dependencies:
   - common/nextflow-forge-iam-policy.yaml
   - common/nextflow-launch-iam-policy.yaml
+  - projects-prod/shared-ce-prod-project.yaml
 
 sceptre_user_data:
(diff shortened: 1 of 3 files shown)


config/projects-prod/mc2-project.yaml
```diff
@@ -53,3 +53,10 @@ template:
 dependencies:
   - common/nextflow-forge-iam-policy.yaml
   - common/nextflow-launch-iam-policy.yaml
+
+sceptre_user_data:
+  ManualComputeEnvs:    # use the shared project's forge batch CEs to run pipelines
```
zaro0508, line 59: this project now depends on `shared-ce-prod-project` so you should append it to the dependencies list, `- projects-prod/shared-ce-prod-project.yaml`.

zaro0508 (changes requested):

zaro0508 (approved):

zaro0508 (approved):

PR #531, opened 2026-03-13
Add support for ignoring projects with ignore flag

zaro0508:
This allows temporarily disabling tower projects without deleting or moving their configuration files, useful for maintenance or troubleshooting specific projects.  It is inline to how sceptre ignores stacks[1]

[1] https://docs.sceptre-project.org/latest/docs/stack_config.html#ignore

depends on #532


thomasyu888 (approved):
nice!

PR #533, opened 2026-03-13
Optimize project loading with caching

zaro0508:
File I/O operations are expensive and doing it multiple times take a performance hit.  Refactor to cache projects configs will not only improve perfomance but also make the code simpler and easier to maintain.

- Before: Project files loaded 3 times during Projects.__init__()
- After: Project files loaded once and cached for subsequent calls
- Eliminates redundant disk I/O and YAML parsing overhead

depends on #531


thomasyu888 (approved):
Nice!

PR #537, opened 2026-03-16
[IT-2361] document the tower project configuration

zaro0508:
Add a AI generated doc to describe the tower project configuration script and supported attributes or schema.

### DIFF
--- a/docs/configure-tower-projects.md
+++ b/docs/configure-tower-projects.md
@@ -1,0 +1,172 @@
+# configure-tower-projects.py
+
+A Python script that synchronizes Seqera Tower (Nextflow Tower) workspaces with
+project configuration files. It creates and manages Tower organizations, workspaces,
+users, compute environments, and credentials.
+
+## Prerequisites
+
+### CloudFormation Stacks
+
+Before running this script, the project configuration files must be deployed as
+CloudFormation stacks using Sceptre. The script reads stack outputs (S3 buckets,
+IAM roles, etc.) that are created by these stacks.
+
+```bash
+# Deploy all project stacks
+pipenv run sceptre launch config/projects-prod
+```
+
+### Python Dependencies
+
+Install dependencies using pipenv:
+
+```bash
+pipenv install
+```
+
+Key dependencies:
+- `boto3` - AWS SDK for Python
+- `pyyaml` - YAML parsing
+- `sagetasks` - Provides `TowerClient` for Tower API interactions
+
+### Environment Variables
+
+The script requires the following environment variables:
+
+| Variable | Description |
+|---|---|
+| `TOWER_ACCESS_TOKEN` | Seqera Tower API access token |
+| `AWS_ACCESS_KEY_ID` | AWS credentials (or use AWS profile/role) |
+| `AWS_SECRET_ACCESS_KEY` | AWS credentials (or use AWS profile/role) |
+| `AWS_DEFAULT_REGION` | AWS region (defaults to `us-east-1`) |
+
+### AWS Permissions
+
+The script requires AWS permissions to:
+- Describe CloudFormation stacks (to retrieve stack outputs)
+- Read Secrets Manager secrets (for Tower Forge credentials)
+
+### Tower Permissions
+
+The Tower access token must have organization admin permissions to:
+- Create/manage workspaces
+- Add/remove participants
+- Create/manage compute environments
+- Create/manage credentials
+
+## Usage
+
+```bash
+pipenv run python bin/configure-tower-projects.py <projects_dir> [options]
+```
+
+### Arguments
+
+| Argument | Description |
+|---|---|
+| `projects_dir` | Path to directory containing project configuration files |
+
+### Options
+
+| Option | Description |
+|---|---|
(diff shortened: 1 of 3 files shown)


docs/configure-tower-projects.md
```diff
@@ -0,0 +1,172 @@
...
+The script uses `CE_VERSION` (i.e. `v13`) to track compute environment versions.
+When updating CE configurations:
+
+1. Increment `CE_VERSION` in the script
+2. Run the script to create new CEs with the updated version
+3. Old CEs (not matching the version) are automatically cleaned up
+
+## Skipping Projects
+
+To temporarily skip a project, add `ignore: true` to its config file:
```
thomasyu888: Consider adding a reason for why someone may want to use this — for example, to temporarily skip building a project when there are long-running jobs with pending infrastructure updates. The caveat is that the user must remember to turn this flag off when they want the workspace to receive the latest upgrades.

docs/configure-tower-projects.md
```diff
@@ -0,0 +1,172 @@
...
+   - Adds users as organization members
+
+2. **Workspace Creation**
+   - Creates a workspace for each project (named after `stack_name`)
+   - Retrieves CloudFormation stack outputs for AWS resource references
+
+3. **User Management**
+   - Extracts user emails from S3 access ARNs in the config
+   - Maps `S3ReadWriteAccessArns` users to Tower `maintain` role
+   - Maps `S3ReadOnlyAccessArns` users to Tower `view` role
```
thomasyu888, line 114: Should we include a note that if all users are mapped to the "view" role, CEs will be deleted? Also, I want to confirm that when deactivating a Tower workspace with shared CEs, the shared CEs are not deleted — that would be a serious issue.

docs/project-config-schema.md
```diff
@@ -0,0 +1,154 @@
...
+
+### `stack_tags` (optional)
+
+AWS resource tags applied to the CloudFormation stack. A `TowerProject` tag is
+automatically added with the `stack_name` value. The `CostCenter` value is
+automatically trimmed to just the program code portion.
+
+```yaml
+stack_tags:
+  Department: IBC
```
thomasyu888:
```suggestion
  Department: TECH
```

docs/project-config-schema.md
```diff
@@ -0,0 +1,154 @@
...
+Custom data passed to the Sceptre template. Currently supports:
+
+#### `ManualComputeEnvs`
+
+List of manual compute environments that reference existing Batch Forge CEs in
+other workspaces. When specified, the workspace uses these manual CEs instead of
+creating its own Batch Forge CEs.
+
+```yaml
+sceptre_user_data:
```
thomasyu888, line 96: Add dependency on shared ce workspace

thomasyu888 (changes requested):
Nice - it's neat to see AI generating documents that pretty much looks good out of the box.

docs/configure-tower-projects.md
zaro0508: > Should we include a note that if all users are mapped to the "view" role, CEs will be deleted?

Done

>  I want to confirm that when deactivating a Tower workspace with shared CEs, the shared CEs are not deleted — that would be a serious issue.

currently deactivating a tower workspace will delete shared CEs.  please add a jira issue describing what you would like changed and how you think it should work.

docs/project-config-schema.md
zaro0508: done

docs/project-config-schema.md
zaro0508: it's added to the `Minimal Example` below.

docs/configure-tower-projects.md
thomasyu888: Ah, thoughts on adding a note like...

> [!CAUTION]
> If you are using shared compute environments, do not deactivate your workspace — doing so will delete the shared compute environment and affect other workspaces that depend on it.

We'll just have to remember during code reviews when people switch from `S3ReadWriteAccessArns` to `S3ReadOnlyAccessArns` that shared CEs aren't accidentally deleted when deactivating worksapces.

I can certainly create a ticket, but I wonder how high priority it is as long as we remember this fact above.

thomasyu888: Or we just completely stop the practice of de-activating workspaces since we have shared CE's now.

docs/configure-tower-projects.md
zaro0508: done

docs/configure-tower-projects.md
zaro0508: added the caution note

> Or we just completely stop the practice of de-activating workspaces since we have shared CE's now.

Good point.  Then the question becomes do you want automated removal of CEs at all?  if so, which conditions should trigger a removal?

thomasyu888: Thats interesting. Ill create a ticket to only remove ces if it's not the shared CE. This way we don't have to always remember.

thomasyu888 (approved):
Thanks!

PR #541, opened 2026-03-17
Protect shared compute environments from cleanup

zaro0508:
Compute environments referenced in ManualComputeEnvs are now excluded from cleanup, preventing accidental deletion of shared CEs that other workspaces depend on.

- Build set of manual CE names and skip them during cleanup
- Update docs to reflect the new cleanup protection behavior
- Remove outdated CAUTION block about shared CE deletion


thomasyu888 (approved):
Nice!

PR #546, opened 2026-04-01
[IT-3968] Create credential using an assumed role

zaro0508:
The creation of a credential was setup to use both an IAM secret keys and role assumption.  We don't need to use both, the preferable option is to use a role.  This change removes the IAM secret key option.

### DIFF
--- a/bin/configure-tower-projects.py
+++ b/bin/configure-tower-projects.py
@@ -618,16 +618,12 @@ class TowerWorkspace:
                 assert cred["provider"] == "aws"
                 assert cred["deleted"] is None
                 return cred["id"]
-        # Otherwise, create a new credentials entry for the project
-        secret_arn = self.stack["TowerForgeServiceUserAccessKeySecretArn"]
-        credentials = self.org.aws.get_secret_value(secret_arn)
+        # Otherwise, create a new credentials entry using role assumption
         data = {
             "credentials": {
                 "name": self.stack_name,
                 "provider": "aws",
                 "keys": {
-                    "accessKey": credentials["aws_access_key_id"],
-                    "secretKey": credentials["aws_secret_access_key"],
                     "assumeRoleArn": self.stack["TowerForgeServiceRoleArn"],
                 },
                 "description": f"Credentials for {self.stack_name}",


bin/configure-tower-projects.py
```diff
@@ -618,16 +618,12 @@ def create_credentials(self) -> int:
...
-        # Otherwise, create a new credentials entry for the project
-        secret_arn = self.stack["TowerForgeServiceUserAccessKeySecretArn"]
-        credentials = self.org.aws.get_secret_value(secret_arn)
+        # Otherwise, create a new credentials entry using role assumption
         data = {
             "credentials": {
                 "name": self.stack_name,
                 "provider": "aws",
                 "keys": {
-                    "accessKey": credentials["aws_access_key_id"],
```
thomasyu888, line 629: Was the access key an ability for people to access s3 data using the AWS cli?

zaro0508: I don't know what why the key was created other than to setup `credentials` in tower.  if keys were passed around to people for accessing S3 then that is a bad practice and it should stop.  A nicely crafted bucket policy is the right approach to allow users to access S3.

thomasyu888: Thanks - will we run into this: https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/pull/312?

zaro0508: hmm, two notes on PR #312:

I believe my comment, `According to the API docs, passing in an assumeRoleArn also requires accessKey and secretKey therefore this will not work`, no longer applies.  Unfortunately I can't find older versions of their API online to verify so i can't confirm.  I do have a [Seqera support ticket](https://support.seqera.io/support/tickets/7346) that does confirms we only need to set assumeRoleArn with the latest version of Tower.  

The `TowerRoleProfile` in #312 is created however it's not applied to anything so nothing gets that profile.  I'm not sure it's even necessary but if it is then it should be applied to the `TowerTask`. 

I think we should try without the TowerRoleProfile.  if that doesn't work then we can add the TowerRoleProfile for another test.

thomasyu888 (approved):

PR #551, opened 2026-04-03
[IT-3968] Replace IAM user credentials with IAM role assumption for Tower Forge

zaro0508:
We are now using a role for tower credentials we no longer need the user access keys. Switch Tower workspace credentials from IAM user access keys to IAM role assumption. The ECS task role (TowerRole) now assumes the per-workspace TowerForgeServiceRole directly via sts:AssumeRole.

This is a redo of PR #547 which failed to deploy with a `Cannot determine region of bucket` error and was reverted in PR #549 

This should also fix issues IT-4104 and IT-4105

depends on #546

### DIFF
--- a/templates/tower-project.j2
+++ b/templates/tower-project.j2
@@ -91,14 +91,6 @@ Conditions:
   SynapseIndexingEnabled: !Equals [!Ref AllowSynapseIndexing, "Enabled"]
 
 Resources:
-  TowerForgeServiceUser:
-    Type: AWS::IAM::User
-
-  TowerForgeServiceUserAccessKey:
-    Type: AWS::IAM::AccessKey
-    Properties:
-      UserName: !Ref TowerForgeServiceUser
-
   TowerForgeServiceRole:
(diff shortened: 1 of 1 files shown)


templates/tower-project.j2
```diff
@@ -111,11 +103,12 @@ Resources:
...
             Action: sts:AssumeRole
             Principal:
-              AWS: !GetAtt TowerForgeServiceUser.Arn
+              AWS: !Ref AccountAdminArns
           - Effect: Allow
             Action: sts:AssumeRole
             Principal:
-              AWS: !Ref AccountAdminArns
+              AWS:
+                'Fn::ImportValue': !Sub '${AWS::Region}-nextflow-ecs-task-definition-TowerRoleArn'
```
zaro0508, line 111: this allow assume role and the `AssumeForgeServiceRole` fixed the `Cannot determine region of bucket` error.

thomasyu888 (approved):
Thanks for helping with this!

PR #557, opened 2026-04-10
[IT-3968] Fix strides ampad deployment

zaro0508:
The project deployments to strides amapad workflow account failed because the TowerRoleArn was moved to nextflow-ecs-task-definition template in PR #552 which broke the reference in tower-project.2 templates.  This attempts to fix the broken reference.

This change parameterizes TowerRoleArn and by default (in prod env) it will reference the Tower role in the account.  If it's deployed to a different account then Tower role ARN needs to be manually looked up and passed in as a literal string.

### DIFF
--- a/config/projects-ampad/agora-project.yaml
+++ b/config/projects-ampad/agora-project.yaml
@@ -23,6 +23,7 @@ parameters:
   TemplateRootUrl: "https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com"
   TowerForgePolicyArn: !stack_output_external nextflow-forge-iam-policy::NextFlowForgePolicyArn
   TowerLaunchPolicyArn: !stack_output_external nextflow-launch-iam-policy::NextFlowLaunchPolicyArn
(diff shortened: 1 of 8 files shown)


thomasyu888 (approved):
Thanks @zaro0508 , the dev environment doesn't test for this, I wonder if there's a way we can do that to prevent issues like this from happening in the future (or if it's worth the effort)

zaro0508:
> Thanks @zaro0508 , the dev environment doesn't test for this, I wonder if there's a way we can do that to prevent issues like this from happening in the future (or if it's worth the effort)

would be nice however i don't how much effort it would be yet.  I think it would be better to just deprecate the whole strides ampad workflow account since we are no longer burdened by the 50 CE limit.

thomasyu888:
> > Thanks @zaro0508 , the dev environment doesn't test for this, I wonder if there's a way we can do that to prevent issues like this from happening in the future (or if it's worth the effort)
> 
> would be nice however i don't how much effort it would be yet. I think it would be better to just deprecate the whole strides ampad workflow account since we are no longer burdened by the 50 CE limit.

Thanks, makes sense as a nice-to-have. The strides ampad workflow account exists not because of the 50 CE limit, but because all workflow execution costs are charged to STRIDES instead of us.

PR #562, opened 2026-04-17
[IT-4104] Replace SMTP with SES and role access

zaro0508:
The latest version of Tower supports using SES directly with IAM role credentials[1].  This means that we can switch from SMTP to SES and eliminate the IAM user, access keys, Lambda, SSM parameters, and the key rotation problem entirely.

Instead of configuring SMTP host/port/user/password, we set Tower to use the AWS SES API via the SDK. The TowerRole (ECS task role) already has IAM credentials available to the containers — we just need to add ses:SendRawEmail permission and configure Tower to use SES directly.

[1] https://docs.seqera.io/platform-enterprise/enterprise/configuration/overview#aws-ses-integration


thomasyu888 (approved):
This is awesome - the thing to test is if we continue to receive an email upon workflow success and/or error

PR #563, opened 2026-04-22
Update Seqera tower version

zaro0508:
Changelogs since existing version:
* Seqera Enterprise v25.3.3 [1]
* Seqera Enterprise v25.3.4 [2]

[1] https://docs.seqera.io/changelog/seqera-enterprise/v25.3.3
[2] https://docs.seqera.io/changelog/seqera-enterprise/v25.3.4

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -3,7 +3,7 @@ profile: {{ var.profile | default() }}
 region: {{ var.region | default("us-east-1") }}
 aws_infra_templates_root_url: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra
 admincentral_cf_bucket: bootstrap-awss3cloudformationbucket-19qromfd235z9
-tower_version: v25.3.1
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
lGTM!

Do upgrades and the SES versioning require us to stop workflows?

zaro0508:
> Do upgrades and the SES versioning require us to stop workflows?

from seqera support..

Going from 25.3.1 to 25.3.4 is a patch series, so the risk is relatively low.

On in-flight workflows: they run as separate jobs on your compute environment of choice, completely separate from the Seqera Platform services. The wokflows themselves won't be affected by the upgrade. During the maintenance window, the platform won't receive progress updates, but workflow status is persisted in the database, and nothing will flip to a wrong terminal state. Once the platform comes back up, events resume, and stats catch up.

thomasyu888 (approved):

PR #580, opened 2026-06-12
filename must end in yaml

thomasyu888:
sagebrain project wasn't being deployed


zaro0508 (changes requested):
actually the account number is wrong on line 24 as well.  it should be the sagebrain account, like in example-sagebrain-project.yaml file.

thomasyu888:
> actually the account number is wrong on line 24 as well. it should be the sagebrain account, like in example-sagebrain-project.yaml file.

@zaro0508 I'm confused by this, I'm seeing the exact same thing, could you help me put up a PR?

zaro0508 (approved):
sorry about that. i'm wrong.  this LGTM.

PR #581, opened 2026-06-16
[IT-5155] Move to seqera enterprise registry

zaro0508:
The legacy registry will remain available for existing major releases and bug-fix updates. However, beginning with Seqera Platform 26.1, all new releases will be published exclusively to the Enterprise Registry. The new registry also contains previous Platform versions, making it a single source for both current and future releases.


thomasyu888 (approved):
LGTM!

PR #595, opened 2026-07-16
[IT-5155] Update seqera tower version (#582)

zaro0508:
Seqera docs says we can do a direct upgrade from v25.3.x [1].

This is a redo of PR #582. Not sure why but v26.1 is no longer available so
we will attempt an update to v26.1.3 

[1] https://docs.seqera.io/platform-enterprise/enterprise/upgrade#upgrading-from-version-253x-to-261


thomasyu888 (approved):

zaro0508:
This upgrade fails to deploy.  need to work with Seqera to figure out the problem.

PR #508, opened 2026-02-24
[IT-2361] Add support for manual CE to reference Batch forge queues

zaro0508:
Implement functionality to create manual compute environments that reference
existing Batch Forge compute environment queues, enabling workspace sharing
of compute resources.

Manual compute environments support both on-demand (single queue) and
spot (separate head/compute queues) Batch Forge configurations.

To enable creation of manual CE instead of forge CE add this to the `sceptre_user_data:` 
block of the tower project configuration file:

```
# Enable manual compute environment by referencing an existing Batch Forge compute environment
sceptre_user_data:
  ManualComputeEnvs:
    - WorkspaceName: shared-ce-project                                 # Source workspace name
      ComputeEnvName: shared-ce-project-on-demand-v1    # Source compute environment name
    - WorkspaceName: shared-ce-project 
      ComputeEnvName: shared-ce-project-spot-v1 
```

How it works:

WorkspaceName: The name of an existing workspace that contains the Batch Forge compute
environment you want to reference

ComputeEnvName: The name of the compute environment in that workspace
(e.g., share-ce-project-spot-v1 for spot instances or shared-ce-project-ondemand-v1
for on-demand)

What happens:

The script will look up the compute environment by name in the source workspace. Extract the AWS Batch queue ARNs (headQueue and computeQueue). Create a new manual compute environment in your workspace that references those same queues. This allows multiple workspaces to share the same underlying AWS Batch compute resources

To disable manual compute environment: Simply omit the ManualComputeEnv section or leave it empty, and the workspace will create its own Batch Forge compute environments as usual.

depends on #511

PR #596, opened 2026-07-17
[IT-5235] Bump CE_VERSION v13 -> v14 to deploy g5 GPU instances

adamjtaylor:
The g5 GPU instance types (g5.xlarge/2xlarge/4xlarge) were added to GPU_EC2_INSTANCE_TYPES in #560, but the deployed v13 compute environments predate that patch (v13 was set in #527, over a month before #560) and CE_VERSION was never bumped. Because check_existing_compute_env() skips any CE whose name already ends in the current CE_VERSION, the live v13 CEs were never re-forged and still lack g5 -- GPU jobs land on the oversized p4d.24xlarge instead of a right-sized single-GPU g5.

Bumping to v14 forces the Batch Forge CEs to be recreated with the current instance-type list (now including g5). Also updates every hardcoded ManualComputeEnvs reference in the project configs so the manual CEs resolve the new v14 source CE names after the old v13 CEs are cleaned up.

depends on #605

### DIFF
--- a/bin/configure-tower-projects.py
+++ b/bin/configure-tower-projects.py
@@ -904,8 +904,6 @@ class TowerWorkspace:
                             "configMode": "Manual",
                             "headQueue": source_ce_head_queue,
                             "computeQueue": source_ce_compute_queue,
-                            "headJobCpus": 8,
-                            "headJobMemoryMb": 15000,
                             "cliPath": "/home/ec2-user/miniconda/bin/aws",
                             "resourceLabelIds": label_ids,
                             "executionRole": source_ce_execution_role,
--- a/bin/configure-tower-projects.py
+++ b/bin/configure-tower-projects.py
@@ -904,6 +904,8 @@ class TowerWorkspace:
                             "configMode": "Manual",
                             "headQueue": source_ce_head_queue,
                             "computeQueue": source_ce_compute_queue,
+                            "headJobCpus": 8,
+                            "headJobMemoryMb": 15000,
                             "cliPath": "/home/ec2-user/miniconda/bin/aws",
                             "resourceLabelIds": label_ids,
                             "executionRole": source_ce_execution_role,
--- a/bin/configure-tower-projects.py
+++ b/bin/configure-tower-projects.py
@@ -15,7 +15,7 @@ import yaml  # type: ignore
 from sagetasks.nextflowtower.client import TowerClient  # type: ignore
 
 # Increment this version when updating compute environments
-CE_VERSION = "v13"
+CE_VERSION = "v14"
 
 REGION = "us-east-1"
 ORG_NAME = "Sage Bionetworks"
--- a/docs/configure-tower-projects.md
+++ b/docs/configure-tower-projects.md
@@ -132,7 +132,7 @@ For each valid project configuration file in the specified directory:
 
 ## Compute Environment Versioning
 
-The script uses `CE_VERSION` (i.e. `v13`) to track compute environment versions.
+The script uses `CE_VERSION` (i.e. `v14`) to track compute environment versions.
 When updating CE configurations:
 
 1. Increment `CE_VERSION` in the script
--- a/docs/configure-tower-projects.md
+++ b/docs/configure-tower-projects.md
@@ -136,8 +136,13 @@ The script uses `CE_VERSION` (i.e. `v14`) to track compute environment versions.
 When updating CE configurations:
 
 1. Increment `CE_VERSION` in the script
-2. Run the script to create new CEs with the updated version
-3. Old CEs (not matching the version) are automatically cleaned up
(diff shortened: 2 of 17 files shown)


adamjtaylor:
### Why this is more than a one-line `CE_VERSION` bump

The [versioning docs](../blob/dev/docs/configure-tower-projects.md#compute-environment-versioning) describe a 3-step process where you only increment `CE_VERSION` in the script (that's all #527 had to do for v12 -> v13). That guidance predates the **Migrate CEs** change (`4fcf0d7`, 2026-03-12), which switched every project to reference the shared workspace's Batch Forge CEs by **hardcoded name** via `ManualComputeEnvs`:

```yaml
sceptre_user_data:
  ManualComputeEnvs:
    - WorkspaceName: shared-ce-prod-project
      ComputeEnvName: shared-ce-prod-project-ondemand-v13
```

So now a version bump must also update those hardcoded `-v13` suffixes. Otherwise the shared workspace forges the new `-v14` CEs, `cleanup_compute_environments()` deletes the old `-v13` ones, and every project's `create_manual_compute_environment()` fails its `get_compute_env_id_by_name(..., "...-v13")` lookup for a source CE that no longer exists.

This PR therefore:
1. bumps `CE_VERSION` -> `v14`,
2. bumps the `ManualComputeEnvs` refs in all 15 project configs,
3. updates the docs to add step 2 so the next bump isn't a footgun.

### Deploy path
Merging here runs the script on the **dev** account (`rw-deploy-projects.yaml` -> `configure-tower-projects.py`) and rebuilds the dev CEs as `v14`. A follow-up **dev -> prod** promotion PR is what actually re-forges the stuck prod GPU CE — which is the real goal here: `g5` has been present in the config on both branches since #560, but the live `v13` CEs were forged before it and were never rebuilt, so GPU jobs fall back to the oversized `p4d.24xlarge`.

zaro0508:
This generally LGTM.  If this isn't urgent then i'de like to hold off on merging.  tower-dev is down right now for debugging  an issue with upgrading seqera tower version (issue IT-5155).

adamjtaylor:
@zaro0508  thank you! Not a "next week" urgent" but by month end would be great.

adamjtaylor:
Folded in a second manual-CE fix that rides the same v14 re-forge:

- `create_manual_compute_environment()` never set `headJobMemoryMb`, so manual CEs fell back to a ~4 GB head node
- mcmicro head-job tasks (SUMMARY_SAMPLESHEET, OMEVALIDATION) request 6 GB → runs died at startup: `Process requirement exceeds available memory -- req: 6 GB; avail: 3.4 GB`
- Forge CEs are fine (hardcode `headJobMemoryMb: 15000`); manual clone dropped it since #508 — not a v13 regression
- Fix: set `headJobCpus: 8` / `headJobMemoryMb: 15000` to match Forge
- Takes effect only on a fresh re-forge (existing manual CEs skipped by name), so it rides this PR's v14 rename — no separate version bump needed

zaro0508:
> Folded in a second manual-CE fix that rides the same v14 re-forge

I would prefer a separate PR for this change.  I would also like to defer to @thomasyu888 for a review of it.

bin/configure-tower-projects.py
```diff
@@ -904,6 +904,8 @@ def create_manual_compute_environment(
                             "configMode": "Manual",
                             "headQueue": source_ce_head_queue,
                             "computeQueue": source_ce_compute_queue,
+                            "headJobCpus": 8,
+                            "headJobMemoryMb": 15000,
```
thomasyu888:
```suggestion
```

@adamjtaylor I'm going to suggest that we shift this into a separate PR for review.

docs/configure-tower-projects.md
```diff
@@ -132,12 +132,17 @@ For each valid project configuration file in the specified directory:
 
 ## Compute Environment Versioning
 
-The script uses `CE_VERSION` (i.e. `v13`) to track compute environment versions.
+The script uses `CE_VERSION` (i.e. `v14`) to track compute environment versions.
```
thomasyu888, line 135: @zaro0508 is this something that is worth automating somehow so that we don't forget to shift the CE version?

thomasyu888 (approved):

docs/configure-tower-projects.md
zaro0508: yes, it would be nice however i'm not sure how to automatically identify when the CE_VERSION should be bumped.

thomasyu888: I think probably whenever there _is_ a new CE version is when it should be bumped

zaro0508:
We need to take special care when deploying a CE_VERSION [bump through the GitHub Action](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/pull/596) because this is considered a migration.  more info at https://sagebionetworks.jira.com/browse/IT-5235?focusedCommentId=337325

PR #605, opened 2026-07-28
fix timing issue when processing tower workspaces

zaro0508:
Workspaces with ManualComputeEnvs reference CEs created by other workspaces. The previous processing order was non-deterministic (dependent on os.walk), which could result in a dependent workspace being processed before its source workspace had created the referenced CE. This sorts workspaces so Batch Forge CE creators run first, ensuring manual CE references resolve correctly during version bumps.

Example: when bumping CE_VERSION from v13 to v14, example-dev-project references shared-ce-dev-project-ondemand-v14 via ManualComputeEnvs. Previously, example-dev-project could be processed first, failing to find the v14 CE that shared-ce-dev-project hadn't created yet. Now shared-ce-dev-project is always processed first since it has no ManualComputeEnvs dependency.

### DIFF
--- a/bin/configure-tower-projects.py
+++ b/bin/configure-tower-projects.py
@@ -1301,7 +1301,8 @@ class TowerOrganization:
             if name in self.manual_compute_env_per_project
         ]
 
-        for name, users in source_projects + dependent_projects:
+        ordered_projects = source_projects + dependent_projects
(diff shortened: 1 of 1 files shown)


adamjtaylor (approved):
Makes sense and LGTM!

